PR #413, opened 2015-01-01
All semver prefixes that start with "v"

sethvargo:
:tada: Happy New Year!

I realize this is a special case, but `rake release` and many other Ruby-related tools automatically prefix a "v" onto tags.

This also raises a more semantic error when a match fails in the `version_composition` (otherwise it's a fun "undefined method :[] for nil").

/cc @schisamo @danielsdeleo

### DIFF
--- a/lib/omnibus/build_version.rb
+++ b/lib/omnibus/build_version.rb
@@ -179,9 +179,9 @@ module Omnibus
     # @return [nil] if no pre-release tag was found
     def prerelease_tag
       prerelease_regex = if commits_since_tag > 0
-                           /^\d+\.\d+\.\d+(?:-|\.)([0-9A-Za-z.-]+)-\d+-g[0-9a-f]+$/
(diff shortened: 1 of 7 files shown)


lamont-granquist:
yeah berks and some other projects use this.  LGTM.

sethvargo:
I don't know why that timeout is happening on Ruby 2.1, but it looks like a fluke to me...

sethvargo:
@lamont-granquist okay. re-ran. It looks like the checksums for ChefDK changed?

lamont-granquist:
i'm not sure what is up with that.

sethvargo:
@lamont-granquist it looks like it was something weird with bzip2. I verified locally and the checksum has, in fact, changed for that version. It's also super super slow to download (even via a browser). Maybe their servers are having problems? I switched to zlib instead and added a progress bar.

Since the tests are now green, I'm gonna merge this in :smile:

PR #414, opened 2015-01-07
Place packager assets into correct location when creating new project

ajkeys:
I noticed that when I created a new project on OSX and edited the resource files in ./resources/pkg my changes were not showing up and omnibus was still using the default included resources. I found that when I moved those resources into ./resources/{my_project_name}/pkg that it picked them up correctly. This pull request correctly generates assets for new projects.

This copies the assets into ./resource/#{name}/ instead of just ./resources. I have also added distribution.xml.erb to be copied on new projects for OSX pkg.

NOTE: I changed the asset paths for all other packagers too as I figured it would be needed for all of them. I have only tested this on OSX when generating a pkg.

### DIFF
--- a/lib/omnibus/generator.rb
+++ b/lib/omnibus/generator.rb
@@ -106,53 +106,53 @@ module Omnibus
     def create_bff_assets
       return unless options[:bff_assets]
 
-      copy_file(resource_path('bff/gen.template.erb'), "#{target}/resources/bff/gen.template.erb")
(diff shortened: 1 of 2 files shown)


schisamo:
Thx @ajkeys! I rebased these commits on current master and merged as 04782102d8e62c348fd85b4c697fe649b827b66a, d794ffadaf49259642d481d8f5bda530f7e35f5a
and 897752b674dba034d6b063db83a1c7e843642d46

PR #416, opened 2015-01-16
Add CPPFLAGS in with_standard_compiler_flags

### LINKED ISSUE #415: Question: why does `with_standard_compiler_flags` not include `CPPFLAGS`?
I've noticed that I'm unable to build `openldap` why Omnibus without including `CPPFLAGS` in the environment, because `openldap` uses `AC_EGREP_CPP` to check for dependency compatibility, e.g. in [`build/configure.m4:1111`](http://www.openldap.org/devel/gitweb.cgi?p=openldap.git;a=blob;f=build/openldap.m4;h=caf8fc2aa6642b3d2ffe02f56dd0a343d4be32f5;hb=HEAD#l1111)

    dnl Check for Cyrus SASL version compatility
    AC_DEFUN([OL_SASL_COMPAT],
    [AC_CACHE_CHECK([Cyrus SASL library version], [ol_cv_sasl_compat],[
            AC_EGREP_CPP(__sasl_compat,[
    #ifdef HAVE_SASL_SASL_H
    #include <sasl/sasl.h>
    #else
    #include <sasl.h>
    #endif
    
    /* Require 2.1.15+ */
    #if SASL_VERSION_MAJOR == 2  && SASL_VERSION_MINOR > 1
            char *__sasl_compat = "2.2+ or better okay (we guess)";
    #elif SASL_VERSION_MAJOR == 2  && SASL_VERSION_MINOR == 1 \
            && SASL_VERSION_STEP >=15
            char *__sasl_compat = "2.1.15+ or better okay";
    #endif
            ],      [ol_cv_sasl_compat=yes], [ol_cv_sasl_compat=no])])
    ])

Setting `env['CPPFLAGS'] = env['CFLAGS']` does the trick, of course, but I'm curious to understand if there is a reason for omnibus to not do this by default?

Cheers,

krallin:
Fixes #415.

Also removed a redundant line in the same function (`CXXFLAGS` was being set to a value that would be ignored down the road).

### DIFF
--- a/lib/omnibus/software.rb
+++ b/lib/omnibus/software.rb
@@ -435,7 +435,6 @@ module Omnibus
             "CC" => "xlc_r -q64",
             "CXX" => "xlC_r -q64",
             "CFLAGS" => "-q64 -I#{install_dir}/embedded/include -D_LARGE_FILES -O",
-            "CXXFLAGS" => "-q64 -I#{install_dir}/embedded/include -D_LARGE_FILES -O",
             "LDFLAGS" => "-q64 -L#{install_dir}/embedded/lib -Wl,-blibpath:#{install_dir}/embedded/lib:/usr/lib:/lib",
             "LD" => "ld -b64",
             "OBJECT_MODE" => "64",
--- a/lib/omnibus/software.rb
+++ b/lib/omnibus/software.rb
@@ -497,8 +497,9 @@ module Omnibus
         # configure scripts which try to be too clever and ignore our explicit
         # CFLAGS and LDFLAGS in favor of pkg-config info
         merge({"PKG_CONFIG_PATH" => "#{install_dir}/embedded/lib/pkgconfig"}).
-        # Set default values for CXXFLAGS.
-        merge('CXXFLAGS' => compiler_flags['CFLAGS'])
+        # Set default values for CXXFLAGS and CPPFLAGS.
+        merge('CXXFLAGS' => compiler_flags['CFLAGS']).
+        merge('CPPFLAGS' => compiler_flags['CFLAGS'])
     end
     expose :with_standard_compiler_flags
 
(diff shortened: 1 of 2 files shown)


krallin:
Quick note: this doesn't appear to be entirely without consequences. Specifically, `ncurses` build from opscode/omnibus-software fails to build, most likely because the wrong header file is getting included due to the double build step.

Removing `CPPFLAGS` from the `env` does fix the issue in that case.

lamont-granquist:
not sure what to do about that.  unclear on how CPPFLAGS incurs a "double build step".   a gist of all the output might help.

krallin:
It's the actual software configuration that builds twice (with two different configurations): https://github.com/opscode/omnibus-software/blob/master/config/software/ncurses.rb

If `CPPFLAGS` is set, then it looks like ncurses ends up picking up its own headers (and this seems to cause conflict).

`CPPFLAGS` was not a problem for any other software recipe I built (~60 included in my project).

Cheers,

lamont-granquist:
Well I'd be okay with a patch to include it by default with another patch to unset it in the ncurses definition -- that seems like it would give better default coverage.  @schisamo ?

schisamo:
@lamont-granquist agreed. @krallin please open a PR in omnibus-software to properly fix the ncurses software definition.

krallin:
@schisamo , @lamont-granquist ,

Finally done here: https://github.com/chef/omnibus-software/pull/359 — sorry for the delay, I was buried under work lately!

Cheers,

schisamo:
Thx @krallin. This was merged as ee3038aba0b24ced578279e8961840217a53f883 and 16f4084ae11d79a309aee3efd0acf6b65f7f76b6.

krallin:
Nice, thanks @schisamo !

PR #418, opened 2015-01-21
Do not memoize current_revision to avoid returning incorrect data

stevendanna:
After a fetch (and other operations), the current revision may
change. Calls to functions such as version_for_cache should reflect
this change.  Memoizing current_revision was previously causing
version_for_cache to return the pre-fetch revision of the software,
leading us to erroneously restore an older revision from the cache.


schisamo:
:+1:

lamont-granquist:
:+1:

sdelano:
:yum:

PR #420, opened 2015-01-30
Improvements for path based builds

stevendanna:
This changeset includes improvements to Digestable, FileSyncer, and PathFetcher.  Most importantly:

- Reset target_shasum in PathFetcher after every fetch, preventing erroneous GitCache misses for builds that produced local artifacts during the previous omnibus run.

More minor improvements:

- Log when an entire directory is being digested.
- Avoid failing when re-syncing read-only files
- Allow users to specify options to the underlying FileSyncer in a Software definition

### DIFF
--- a/lib/omnibus/digestable.rb
+++ b/lib/omnibus/digestable.rb
@@ -63,7 +63,7 @@ module Omnibus
     #
     def digest_directory(path, type = :md5)
       digest = digest_from_type(type)
-
+      log.info(log_key) { "Digesting #{path} with #{type}" } if defined?(log) && defined?(log_key)
       FileSyncer.glob("#{path}/**/*").each do |filename|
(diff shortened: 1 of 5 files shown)


lib/omnibus/digestable.rb
```diff
@@ -63,7 +63,7 @@ def digest(path, type = :md5)
     #
     def digest_directory(path, type = :md5)
       digest = digest_from_type(type)
-
+      log.info(log_key) { "Digesting #{path} with #{type}" } if defined?(log) && defined?(log_key)
```
schisamo: I think we can avoid the guard by just mixing in the `Logging` module.

stevendanna:
Removed the defined? guards.

schisamo:
LGTM. :shipit:

PR #421, opened 2015-02-02
Impove docs regarding git-cache sharing

krallin:
Sharing the git cache across build slaves has a few caveats (though it's indeed possible). 

Thought I'd share a bit of documentation after debugging my way through it!


yzl:
+1

schisamo:
Thx! This was merged as 443d44cb4787d59d562b4df2ee8237af345dc2df.

PR #429, opened 2015-02-12
Bump fauxhai to 2.3, required for current rspec tests

stevendanna:
Omnibus's spec tests requires fauxhai platform data from SuSE 12.0 and
Debian 7.6, which is only available in a newer version.


schisamo:
:shipit:

PR #430, opened 2015-02-13
Introduce version-manifest to omnibus

stevendanna:
This changeset introduces the following features:

  - Omnibus creates a version manifest in a text format (intended
    for humans) without depending on omnibus-software. 

  - Omnibus creates a version manifest in a JSON format (intended
    for machines). 

  - Omnibus can ingest a user-provided, JSON-formatted version
    manifest during the build process. The JSON manifest will be
    used to lock the software sources and version used when
    fetching. Please see the following sections for more details.

* Manifest Ingestion

Ingesting an externally created manifest is NOT intended to be the
standard mode of operation for an omnibus built project. Most projects
should continue to use the `default_version` method of software
definitions and the `override` method of project definition to control
which version of a software get built.

This feature is included to support improvements to the build
pipelines for opscode-omnibus (Chef Server) and other Chef Software,
Inc managed projects by:

  - allowing the opscode-omnibus software to use floating constraints,
    ensuring new features in API services are immediately integrated
    into the Chef Server package.

while still

  - allowing users to rebuild a specific version of an omnibus built
    project from source.

  - ensuring all versions of all software in a given build are captured
    in a format that can be commited to source control.

  - providing a possible base for automated generation of other
    release-related artifacts such as change logs.

Chef Software, Inc reviewers should note that the manifest will NOT be
ingested at any point in the standard build pipeline.

* Text Manifest Format

For backwards compatibility, the text version manifest is produced in
exactly the same manner as the previous version manifest produced by
omnibus-software.

* JSON Manifest Format v1

The first version of the manifest format takes the following form:

```json
{
    "manifest_format" : MANIFEST_FORMAT_NUMBER,
    "software" : {
    "SOFTWARE_NAME" : {
      "described_version": STRING(see below)
      "locked_version": STRING(see below)
      "locked_source": HASH(see below)
      "source_type": "path"|"url"|"git"|"project_local"
    }
}
```

  - `described_version` is the user-specified version of the software
    (via default_version or an override) at the time of the build.

  - `locked_version` is the calculated version of the source artifact
    that was fetched. For git sources this will be a git ref and may
    differ from the described_version.

  - `locked_source` is the source parameters used to fetch the source
    artifact.

  - `source_type` specifies the type of source fetcher that was used.

Only non-null fields are present in the manifest.  Software with a
`project_local` source type may not have a source or either version
field. Software with a `path` source type may not have a either
version field.

* Build Reproducibility

When ingesting a manifest, omnibus will do the following:

- Use the `locked_version` and `locked_source` when fetching the build
artifact.

- Proceed with the usual omnibus build procedures

This procedure does NOT guarantee that a given build will produce a
functionally equivalent artifact.  This is especially true for

- projects using path or project_local based software

- projects using software where the build steps also pull down new
software (bundle install)

- projects using software where the build steps produce different
output depending on environmental factors (time, state of the build
machine, etc)

### DIFF
--- a/lib/omnibus/cli.rb
+++ b/lib/omnibus/cli.rb
@@ -71,20 +71,20 @@ module Omnibus
       default: nil
     desc 'build PROJECT', 'Build the given Omnibus project'
     def build(name)
-      project = Project.load(name)
+      manifest = if @options[:use_manifest]
+                   Omnibus::Manifest.from_file(@options[:use_manifest])
+                 else
+                   nil
+                 end
 
+      project = Project.load(name, manifest)
       say("Building #{project.name} #{project.build_version}...")
-
-      if @options[:use_manifest]
-        project.manifest(Omnibus::Manifest.from_file(@options[:use_manifest]))
-      end
-
       project.download
       project.build
 
       if @options[:output_manifest]
         File.open('version-manifest.json', 'w') do |f|
-          f.write(JSON.pretty_generate(project.manifest.to_hash))
+          f.write(JSON.pretty_generate(project.built_manifest.to_hash))
         end
       end
     end
--- a/lib/omnibus/cli.rb
+++ b/lib/omnibus/cli.rb
@@ -61,12 +61,32 @@ module Omnibus
     #
     #   $ omnibus build chefdk
     #
+    method_option :output_manifest,
+      desc: "Create version-manifest.json in current directory at the end of the build",
+      type: :boolean,
+      default: false
+    method_option :use_manifest,
+      desc: "Use the given manifest when downloading software sources.",
+      type: :string,
+      default: nil
     desc 'build PROJECT', 'Build the given Omnibus project'
     def build(name)
       project = Project.load(name)
 
       say("Building #{project.name} #{project.build_version}...")
-      project.build_me
+
+      if @options[:use_manifest]
+        project.manifest(Omnibus::Manifest.from_file(@options[:use_manifest]))
+      end
+
+      project.download
+      project.build
+
+      if @options[:output_manifest]
+        File.open('version-manifest.json', 'w') do |f|
+          f.write(JSON.pretty_generate(project.manifest.to_hash))
+        end
+      end
     end
 
     #
--- a/lib/omnibus/project.rb
+++ b/lib/omnibus/project.rb
@@ -17,6 +17,9 @@
 
 require 'time'
 require 'json'
+require 'omnibus/manifest'
+require 'omnibus/manifest_entry'
+require 'omnibus/reports'
 
 module Omnibus
   class Project
@@ -668,6 +671,40 @@ module Omnibus
     expose :ohai
 
     #
+    # Write a json-format version manifest to the specified location
+    # at the end of the build. If no path is specified
+    # +install_dir+/version-manifest.json is used.
+    #
+    # @example
+    #   with_json_manifest
+    #
+    # @return [String]
+    #
+    def with_json_manifest(path=File.join(install_dir, "version-manifest.json"))
+      @write_json_manifest=true
+      @json_manifest_path=path
+    end
+    expose :with_json_manifest
+
+    #
+    # Write a text-formatted manifest to the specified location
+    #(+install_dir+/version-manifest.txt if none is provided).
+    #
+    # This manifest uses the same format used by the
+    # 'version-manifest' software definition in omnibus-software.
+    #
+    # @example
+    #   with_text_manifest
+    #
+    # @return [String]
+    #
+    def with_text_manifest(path=File.join(install_dir, "version-manifest.txt"))
+      @write_text_manifest=true
+      @text_manifest_path=path
+    end
+    expose :with_text_manifest
+
+    #
     # @!endgroup
     # --------------------------------------------------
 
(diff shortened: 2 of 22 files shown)


seth:
Can we merge this now :troll:

lib/omnibus/project.rb
```diff
@@ -904,38 +904,61 @@ def <=>(other)
...
-
-      # Compress
       compress_me
     end
 
     #
+    # Download and build the project. Preserved for backwards
+    # compatibility.
+    #
+    def build_me(manifest=resolve_manifest)
```
lamont-granquist: can this be a lazily allocated instance accessor instead?

```
def manifest
  @manifest ||= resolve_manifest
end
```

lib/omnibus/software.rb
```diff
@@ -581,6 +582,14 @@ def builder
       @builder ||= Builder.new(self)
     end
 
+    def manifest_entry
+      Omnibus::ManifestEntry.new(name, {
+                                   source_type: source_type,
+                                   described_version: version,
+                                   locked_version: fetcher.resolved_version,
+                                   locked_source: source})
```
schisamo: A couple of thoughts here:

* Since we are converting to a manifest entry maybe a method name of `to_manifest_entry` makes sense.
* Should we cache this?
* Setting attributes/properties via a hash in the initializer could prove error prone. Why don't we just set values using the accessor?

```ruby
def manifest_entry
  @manifest_entry ||= begin
    manifest_entry = Omnibus::ManifestEntry.new(name)
    manifest_entry.source_type = source_type
    manifest_entry.described_version = version
    manifest_entry.locked_version = fetcher.resolved_version
    manifest_entry.locked_source = source
    manifest_entry
  end
end
```

lib/omnibus/cli.rb
```diff
@@ -66,7 +66,9 @@ def build(name)
       project = Project.load(name)
 
       say("Building #{project.name} #{project.build_version}...")
-      project.build_me
+      manifest = project.resolve_manifest
+      project.download(manifest)
```
schisamo: It feels like a project instance should just expose a `#manifest` instance method and use that internally in the `#download` instance method.

lib/omnibus/project.rb
schisamo: :+1: See https://github.com/chef/omnibus/pull/430/files#r24699183

lib/omnibus/software.rb
```diff
@@ -693,16 +703,34 @@ def version_for_cache
...
+                     NullFetcher.new(self)
+                   end
+    end
+
+    #
+    # The type of source specified for this software defintion.
+    #
+    # @return [String]
+    #
+    def source_type
```
schisamo: This feels like it is an attribute of the fetcher: `some_software_instance.fetcher.type`.

schisamo:
@stevendanna Excited about this feature, overall a great start! Let's be sure to add some functional test coverage also.

lib/omnibus/software.rb
stevendanna: @schisamo I've renamed this to_manifest_entry.  On the other two points, here are my thoughts, would love to hear what you think:

1) Re caching: `to_manifest_entry` should return a manifest entry for the Omnibus::Software in its current state, which would complicate caching.  Namely, if we were to cache this, would we then invalidate the caching anytime the source or version was changed on the Omnibus::Software?  Both sides of this are a bit theoretical at the moment since, in a typical run, this will only be called once since we do cache the entire manifest at the project level.

2) I'd like to think of Omnibus::ManifestEntry as a sort of "Type/Value object" that shouldn't change after creation.  This is why I've set everything we care about at instantiation and then only provided reader methods.  Any thoughts on the Ruby-way of doing this?

schisamo: @stevendanna Your responses make total sense... :+1:.

schisamo:
@stevendanna This is looking good. A couple of final questions for you:

1. In the name of keeping things simple: Should we just make both the TXT and JSON manifests mandatory and always generate them at your suggested default paths (e.g. `INSTALL_DIR/version-manifest.*`)? I'm not sure we want to make something like this configurable. I also don't see a downside to all Omnibus projects gaining this useful feature.
2. Can we add functional test coverage for `Omnibus::NetFetcher` and `Omnibus::PathFetcher`? I want to make sure we don't incur any future regressions on this feature.

danielsdeleo:
Awesome. One thing I'd recommend to change. At the time you load the project you can know whether you're going to be ingesting a manifest. So you could pass it to the Project's constructor, and then to individual software constructors, which would remove a lot of switch-y behavior from the interfaces of these things. Specifically, `Fetcher#use_manifest_entry`, `Project#manifest`, `Software#fetch` each involve an outsider telling some existing object to flip a bit to a very different mode of operation. I like to keep stuff like this in the constructor to ensure that the behavior changes are more strictly contained in each object. Also it's less likely you could cause an ordering issue since each object knows from the time it's created how it's supposed to behave.

stevendanna:
@danielsdeleo I've just pushed a change that hopefully moves this closer towards the changes you recommended.  Currently, the only real "decision point" where there is a change between a user-provided manifest and a self-generated manifest is here:

https://github.com/chef/omnibus/pull/430/files#diff-2753487ba152ae43f78b4da0c96b8e44R102

That is hard to do in the constructor itself without more major refactoring.  A software just doesn't know enough about itself to create a manifest entry until it evaluates the software definition on disk.

A nice impact of this change is that the fetchers are now completely ignorant of the manifest.  They take a manifest-like thing in their constructor, but then just assume that the information they received in their constructor is absolutely correct.  Any symbolic-version -> real-version resolution should be preformed by the user (i.e. Omnibus::Software) using the helper method Fetcher.resolve_version.

@schisamo With the new design above, the functional tests moved around a bit. The fetchers don't care too much about the manifest, but I did add tests to ensure the #resolve_version function works for each class and some top level Fetcher tests.

Regarding making with_text_manifest and with_json_manifest the defaults, I have no objection as long as we are OK with changing the build output people will get on upgrade.

sdelano:
This LGTM. It took a bit of spelunking to detect whether the ordering changed much, but this is a good step in the right direction. I'm :+1: x 1000 for trying this out on the Chef Server.

schisamo:
@stevendanna Let's just make the generation of `version-manifest.txt` and `version-manifest.json` the default behavior. It simplifies the code (and projects opting into use it) and fits in with Omnibus's opinionated approach. These manifests are useful and there is really no reason you wouldn't want them!

We'll just treat this as a new feature which shouldn't break anyone in a backward compatible way.

danielsdeleo:
:+1:

lib/omnibus/cli.rb
```diff
@@ -61,12 +61,32 @@ def execute!
...
       say("Building #{project.name} #{project.build_version}...")
-      project.build_me
+      project.download
+      project.build
+
+      if @options[:output_manifest]
+        File.open('version-manifest.json', 'w') do |f|
+          f.write(JSON.pretty_generate(project.built_manifest.to_hash))
+        end
+      end
```
schisamo: @stevendanna Should we be generating a 2nd manifest here? It would be safer to either drop this option and copy the one generated by `Omniubs::Project#build` into place. This would work similar to what we do with packages:
https://github.com/chef/omnibus/blob/master/lib/omnibus/project.rb#L1046-L1049

I have some mixed feelings about this approach but at least things would be consistent.

stevendanna: It is a bit round-about, but built_manifest shouldn't recalculate anything.  Project#built_manifest calls Software.manifest_entry

https://github.com/chef/omnibus/pull/430/files#diff-9d12b132bef62277d2011f5c7bf2368cR970

Software#manifest_entry is memoized at the time of creation:

https://github.com/chef/omnibus/pull/430/files#diff-2753487ba152ae43f78b4da0c96b8e44R103

So it will always return either the manifest entry it was passed on instantiation or the one it generated itself during the build.

PR #434, opened 2015-02-20
Allow solaris to use mapfiles

scotthain:
This change allows the optional use of a Solaris 10 linker mapfile. It also updates the whitelist for Solaris 10 library healthcheck to enable libraries that are 10+.

This also has 1 breaking change. ```solaris_compiler``` has been removed.

@chef/ociv @lamont-granquist @benr


lib/omnibus/config.rb
```diff
@@ -432,10 +432,16 @@ def reset!
...
+    # see http://docs.oracle.com/cd/E23824_01/html/819-0690/chapter5-1.html
+    # Path is relative to the 'files' directory in your omnibus project
+    #
+    # For example:
+    #
+    #     /PATH/files/my_map_file 
     #
     # @return [String, nil]
-    default(:solaris_compiler, nil)
+    default(:solaris_linker_mapfile, nil)
```
schisamo: To follow idioms we used in other config values, let's just set a default value here and then check for it's existence when it's used in `Omnibus::Software#with_standard_compiler_flags `:

```ruby
# Solaris linker mapfile to use during compilation. This is relative {#project_root}.
#
# @see http://docs.oracle.com/cd/E23824_01/html/819-0690/chapter5-1.html
# @return [String]
default(:solaris_linker_mapfile, 'files/mapfile/solaris')
```

lib/omnibus/software.rb
```diff
@@ -483,13 +483,26 @@ def with_standard_compiler_flags(env = {}, opts = {})
...
+
+      if solaris2?
+        # in order to provide compatibility for earlier versions of libc on solaris 10,
+        # we need to specify a mapfile that restricts the version of system libraries
+        # used. See http://docs.oracle.com/cd/E23824_01/html/819-0690/chapter5-1.html
+        # for more information
+        # use the mapfile if it is configured, otherwise ignore it
+        unless Config.solaris_linker_mapfile.nil?
+         mapfile = " -M #{project.files_path}/#{Config.solaris_linker_mapfile}"
+        end
```
schisamo: Per my other comment this can now change to:

```ruby
ld_options = "-R#{install_dir}/embedded/lib"
mapfile_path = File.expand_path(Config.solaris_linker_mapfile, Config.project_root)
ld_options  << " -M #{mapfile_path}" if File.exist?(mapfile_path)

extra_linker_flags.merge!(
  "LD_OPTIONS" => ld_options
)
```

PR #438, opened 2015-03-03
Add `omnibus manifest PROJ` command

stevendanna:
This command allows you to generate a manifest without doing a
complete build.  It can be useful for diagnostics or for getting a
sense of what is going to change on the next build.

### DIFF
--- a/lib/omnibus/cli.rb
+++ b/lib/omnibus/cli.rb
@@ -90,6 +90,16 @@ module Omnibus
     end
 
     #
+    # Generate a version manifest for the given project definition
+    #
+    #   $ omnibus manifest PROJECT
+    #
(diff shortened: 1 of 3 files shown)


lib/omnibus/cli.rb
```diff
@@ -90,6 +90,16 @@ def build(name)
     end
 
     #
+    # Generate a version manifest for the given project definition
+    #
+    #   # omnibus manifest PROJECT
```
schisamo: I think that second `#

schisamo:
Can we ensure the README's (main project one and template in project generator) mention the new command?

stevendanna:
@schisamo +1 on both comments, next push will fix them.

schisamo:
:shipit:

PR #439, opened 2015-03-03
Add `omnibus generate changelog` command

stevendanna:
ChangeLog-Entry: A new `omnibus generate changelog` command generates
an opinionated CHANGELOG entry based on metadata in commit messages.

### DIFF
--- a/lib/omnibus/semantic_version.rb
+++ b/lib/omnibus/semantic_version.rb
@@ -1,26 +1,53 @@
+#
+# Copyright 2015 Chef Software, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'mixlib/versioning'
+
 module Omnibus
   class SemanticVersion
     def initialize(version_string)
-      @has_v = !!(version_string =~ /^v/)
-      @version = Gem::Version.new(version_string.gsub(/^v/, ""))
+      @prefix = if version_string =~ /^v/
+                  "v"
+                else
+                  ""
+                end
+      @version = Mixlib::Versioning.parse(version_string.gsub(/^v/, ""))
+    end
+
+    def to_s
+      "#{prefix}#{version}"
     end
 
     def next_patch
-      major, minor, patch = @version.segments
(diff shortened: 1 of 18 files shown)


lib/omnibus/semantic_version.rb
```diff
@@ -0,0 +1,26 @@
+module Omnibus
+  class SemanticVersion
+    def initialize(version_string)
+      @has_v = !!(version_string =~ /^v/)
+      @version = Gem::Version.new(version_string.gsub(/^v/, ""))
```
schisamo: Pretty sure this will choke on SemVer-compliant pre-release and build versions. This would probably be a good time to begin using [mixlib-versioning](https://github.com/chef/mixlib-versioning) for Omnibus's internal version parsing.

schisamo:
This looks like an awesome start @stevendanna! Super excited for this feature.

lib/omnibus/semantic_version.rb
stevendanna: I *knew* we had a library for this.  Thanks!

stevendanna:
TODO:

- [ ] Wrap my head around cucumber so we can write some basic tests for the actual output of the `omnibus changelog generate` command`
= [ ] Bubble up ChangeLogs from components

tylercloke:
LGTM

lib/omnibus/semantic_version.rb
```diff
@@ -0,0 +1,53 @@
...
+
+module Omnibus
+  class SemanticVersion
+    def initialize(version_string)
+      @prefix = if version_string =~ /^v/
+                  "v"
+                else
+                  ""
+                end
+      @version = Mixlib::Versioning.parse(version_string.gsub(/^v/, ""))
```
schisamo: What happens if a version string does not adhere to a format `Mixlib::Versioning` can parse? Can we gracefully throw an exception and add some test coverage for that case?

stevendanna: Done.

stevendanna:
@chef/lob @schisamo Are we good for an initial merge?  As this is an experimental feature, this code will likely be in flux over the coming weeks, but it won't affect existing omnibus functionality.  We have other changes that we'd like to make but I'd prefer to do those incrementally rather than keeping this in a branch for a long time and constantly rebasing.

lib/omnibus/semantic_version.rb
```diff
@@ -0,0 +1,58 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'mixlib/versioning'
+
+module Omnibus
+  class SemanticVersion
+    class InvalidVersion < StandardError; end
```
schisamo: We should throw this in `lib/omnibus/execeptions.rb` and also follow the idioms for other exception classes.

schisamo:
Modulo my final comments about `InvalidVersion` I'm plus one on getting this merged.

stevendanna:
Exception updated and rebased

schisamo:
:shipit:

PR #440, opened 2015-03-03
Allow the use of SHA checksums in source definitions

szechyjs:
Not all software providers post MD5 checksums for their packages, most seem to favor SHA. This PR adds the ability to use SHA1, SHA256, or SHA512 in the source definition.

Some work may be required to DRY up the tests, unfortunately I'm not an expert with RSpec.

### DIFF
--- a/lib/omnibus/fetchers/net_fetcher.rb
+++ b/lib/omnibus/fetchers/net_fetcher.rb
@@ -26,6 +26,9 @@ module Omnibus
     # tar probably has compression scheme linked in, otherwise for tarballs
     TAR_EXTENSIONS = %w(.tar .tar.gz .tgz .bz2 .tar.xz .txz)
 
+    # Digest types used for verifying file checksums
+    DIGESTS = [:md5, :sha1, :sha256, :sha512]
+
     #
     # A fetch is required if the downloaded_file (such as a tarball) does not
     # exist on disk, or if the checksum of the downloaded file is different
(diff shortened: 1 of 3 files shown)


lib/omnibus/fetchers/net_fetcher.rb
```diff
@@ -215,6 +218,17 @@ def extract
     end
 
     #
+    # The digest type defined in the software definition
+    #
+    # @return [Symbol]
+    #
+    def digest_type
+      DIGESTS.each do |digest|
```
stevendanna: Currently, the DIGESTS array is listed with the weakest hash function first.  I'm wondering if we should reverse the order so that the strongest hash function is preferred.

stevendanna:
One minor comment, but overall I'm :+1: on this feature.

lib/omnibus/fetchers/net_fetcher.rb
szechyjs: I thought the same thing. I'll reverse the order.

stevendanna:
:+1: Thanks for the update.

szechyjs:
Bump!

@schisamo @scotthain @yzl

scotthain:
This looks good to me, thanks!

schisamo:
@szechyjs This is great. Let's rebase this branch onto master then I'll get it merged.

szechyjs:
rebased.

PR #445, opened 2015-03-09
Use File.expand_path in the pkg packagers

stevendanna:
When building a project with the base_dir set to "./local" for test
builds, the pkg packager will fail since the underlying command
doesn't understand relative paths.

### DIFF
--- a/lib/omnibus/project.rb
+++ b/lib/omnibus/project.rb
@@ -179,7 +179,7 @@ module Omnibus
       if null?(val)
         @install_dir || raise(MissingRequiredAttribute.new(self, :install_dir, "/opt/chef"))
       else
-        @install_dir = val.tr('\\', "/").squeeze("/").chomp("/")
+        @install_dir = File.expand_path(val.tr('\\', "/").squeeze("/").chomp("/"))
       end
     end
     expose :install_dir
--- a/spec/support/ohai_helpers.rb
+++ b/spec/support/ohai_helpers.rb
@@ -11,10 +11,18 @@ module Omnibus
         ohai = Mash.from_hash(Fauxhai.mock(options, &block).data)
         allow(Ohai).to receive(:ohai).and_return(ohai)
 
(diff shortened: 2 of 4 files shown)


stevendanna:
I've been carting around this change for a while.

lib/omnibus/packagers/pkg.rb
```diff
@@ -192,8 +192,8 @@ def build_component_pkg
           --identifier "#{safe_identifier}" \\
           --version "#{safe_version}" \\
           --scripts "#{scripts_dir}" \\
-          --root "#{project.install_dir}" \\
-          --install-location "#{project.install_dir}" \\
+          --root "#{::File.expand_path(project.install_dir)}" \\
```
schisamo: We shouldn't need the `::` here as a `Omnibus::Packagers::File` doesn't exists.

schisamo:
@stevendanna Is there some way to apply this expansion one time in `Omnibus::Project#install_dir`? Otherwise it seems we'll need to perform path expansion in every class that utilizes `Omnibus::Project#install_dir`.

stevendanna:
@schisamo Yup, good call.  I threw this up here too fast as I wanted to make sure I stopped forgetting about it.  I'll try to expand path at a more central location.

stevendanna:
The changes to the Windows tests have to do with the fact that the behavior of expand_path is platform dependent, so the windows tests were failing on my OS X machine.  Gonna consider this on hold until I have some time to test this on windows.

schisamo:
@stevendanna Perhaps we could hook the the project up to Appveyor?

ksubrama:
It's now looked into appveyor. @stevendanna  - you can scratch that itch now if you want.

schisamo:
@stevendanna Mind getting this branch rebased on latest master?

stevendanna:
I have no idea if this is relevant anymore and we haven't needed it. So, closing.

PR #446, opened 2015-03-10
Respect append_timestamp config when constructing a version number

stevendanna:
This helper method allows users to specify a version as:

build_version Omnibus::BuildVersion.static_with_timestamp("1.3.4")

which will append a timestamp to the build version.  This may be
helpful for build pipelines in which you may have multiple builds with
the same, unreleased, user-supplied version.

### DIFF
--- a/spec/unit/build_version_dsl_spec.rb
+++ b/spec/unit/build_version_dsl_spec.rb
@@ -7,8 +7,9 @@ module Omnibus
 
     let(:version_string) { "1.0.0" }
     let(:description) { nil }
+    let(:today_string) { Time.now.utc.strftime(Omnibus::BuildVersion::TIMESTAMP_FORMAT) }
 
-    let(:zoo_version) { double("BuildVersion", semver: "5.5.5", custom: "7.7.7") }
+    let(:zoo_version) { double("BuildVersion", semver: "5.5.5", custom: "7.7.7", build_start_time: today_string) }
     let(:zoo_software) { double("software", name: 'zoo', project_dir: '/etc/zoo', version: "6.6.6") }
 
     describe "when given nil" do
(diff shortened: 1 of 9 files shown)


schisamo:
@stevendanna Would it be easier to just force all build versions to respect [Omnibus::Config. append_timestamp](https://github.com/chef/omnibus/blob/8ec7b37acd027416aee29b95370dcf7e309316b6/lib/omnibus/config.rb#L454-L457)? That config value's stated intent is:

> Append the current timestamp to the version identifier.

stevendanna:
@schisamo Happy to do that, but just as a heads up, that is inevitably going to change the current behavior for some user so we should bump the version numbers appropriately.

stevendanna:
A data point on "how much" behavior change it will cause:

32 unit test failures from respecting that option :)

stevendanna:
@schisamo I've updated this to more generally respect append_timestamp.  The `has_timestamp?` function is there to prevent us from adding an extra timestamp to versions that are already in "semver" format as a result of being created by Omnibus::BuildVersion.semver.  Since the auto-generated template uses such a setup, my hope is that most external users won't need to change much.  However, some users may still need to update their config to set append_timestamp to false.

stevendanna:
cc @chef/lob

spec/unit/compressors/dmg_spec.rb
```diff
@@ -3,6 +3,7 @@
 module Omnibus
   describe Compressor::DMG do
     let(:project) do
+      Config.append_timestamp(false)
```
schisamo: It might be easier to set `Config.append_timestamp(false)` option globally here:
https://github.com/chef/omnibus/blob/398b1d8b59267ceec2506b53e7e66eb5aacfe9cf/spec/spec_helper.rb#L58

This would save having to change all these specs.

spec/unit/build_version_dsl_spec.rb
```diff
@@ -2,13 +2,15 @@
 
 module Omnibus
   describe BuildVersionDSL do
+    before { Config.append_timestamp(false) }
     let(:subject_with_version) { described_class.new(version_string) }
     let(:subject_with_description) { described_class.new(&description) }
 
     let(:version_string) { "1.0.0" }
     let(:description) { nil }
+    let(:today_string) { Time.now.utc.strftime('%Y%m%d') }
```
schisamo: If you fully format `today_string` with `Omnibus::BuildVersion::TIMESTAMP_FORMAT` you can use it in all the matches below: https://gist.github.com/schisamo/4c03f23c23ce0139d5ca

schisamo:
Overall this is looking great @stevendanna. I know we haven't done this with the last few PRs but let's update the `CHANGELOG.md` with a description of this feature. We can also call out to people they can disable this default behavior by change setting `append_timestamp` in their `omnibus.rb` OR using `--override append_timestamp:false` on the commandline. We should document that in the `README` and `CHANGELOG`.

schisamo:
:+1: let's :shipit:

PR #452, opened 2015-03-17
Cache Builder#shasum before Builder#build to ensure consistent result

### LINKED ISSUE #450: Dynamic Build Steps breaks Git Caching
TL;DR: Dynamically defined build steps are can cause issues with git caching in omnibus. We're currently experiencing this when building Postgres, where the second build invalidates the cache and we have to start over from a much earlier point in the build than we should have to.

From a discussion with @stevendanna earlier:

```
The shasum of the software includes a digest of the builder:

  https://github.com/chef/omnibus/blob/master/lib/omnibus/software.rb#L821

The digest of the builder, includes a digest of the collected build_commands:

  https://github.com/chef/omnibus/blob/master/lib/omnibus/builder.rb#L614-L616

Part of our postgresql definition looks like this:

  # Postgres 9.2 is our "real" Postgres installation (prior versions
  # that are installed are solely to facilitate upgrades).  As a
  # result, we need to have the binaries for this version available
  # with the other binaries used by Private Chef.
  block do
    Dir.glob("#{install_dir}/embedded/postgresql/9.2/bin/*").sort.each do |bin|
      link bin, "#{install_dir}/embedded/bin/#{File.basename(bin)}"
    end
  end

Note that both `block` and `link` are part of the builder DSL.
If we calculate the shasum of the software *before* the build,
the build_commands are part of the digested:

[Builder: postgresql92] I | Incremental shasum(Execute: `./configure --prefix=/opt/postgresqltest/embedded/postgresql/9.2 --.....
[Builder: postgresql92] I | Incremental shasum(Execute: `make world -j 3') for builder: 562b3d29700088e3cdd140fb812c36a776f1ed63f8a3a8fd625a71856fc2ec50
[Builder: postgresql92] I | Incremental shasum(Execute: `make install-world -j 3') for builder: ff6b1ed515a24c5497365f82d21a41eb6a9add0aa2157bb48a25745e38707a8b
[Builder: postgresql92] I | Incremental shasum(<Dynamic Ruby block>) for builder: 842bdd5519e0f2020e5a69e8d0f4b218bbc88e1ec8e80306047d07258a96ba15

However, if we calculate the shasum *after* the build, the
dynamic resource block has been executed and all of those link
statements have been added to the build_commands:

[Builder: postgresql92] I | Incremental shasum(Execute: `./configure --prefix=/opt/postgresqltest/embedded/postgresql/9.2 --with-.....
[Builder: postgresql92] I | Incremental shasum(Execute: `make world -j 3') for builder: 562b3d29700088e3cdd140fb812c36a776f1ed63f8a3a8fd625a71856fc2ec50
[Builder: postgresql92] I | Incremental shasum(Execute: `make install-world -j 3') for builder: ff6b1ed515a24c5497365f82d21a41eb6a9add0aa2157bb48a25745e38707a8b
[Builder: postgresql92] I | Incremental shasum(<Dynamic Ruby block>) for builder: 842bdd5519e0f2020e5a69e8d0f4b218bbc88e1ec8e80306047d07258a96ba15
[Builder: postgresql92] I | Incremental shasum(link `/opt/postgresqltest/embedded/postgresql/9.2/bin/clusterdb' to `/opt/postgresqltes...
[Builder: postgresql92] I | Incremental shasum(link `/opt/postgresqltest/embedded/postgresql/9.2/bin/createlang' to `/opt/...
[Builder: postgresql92] I | Incremental shasum(link `/opt/postgresqltest/embedded/postgresql/9.2/bin/....
[Builder: postgresql92] I | Incremental shasum(link `/opt/postgresqltest/embedded/postgresql/9.2/bin/dropdb' to `/opt/postgresqltest/embe..

Thus, the following can happen:

1) First run, something *before* postgresql dirties the
cache. This causes us to start building before we ever call
software#shasum:

https://github.com/chef/omnibus/blob/master/lib/omnibus/software.rb#L763-L767

2) At the end of the build, we update the gitcache, calling
shasum for the first time and building the digest with all the link resources as
shown above.

3) On the second run, we should be able to restore from the
cache, but we call software#shasum before the build, which means
the link resources haven't been added to the build_commands and
thus we calculate a different shasum.
```

There are a couple of ways that we can approach a fix for this:

1. We could convert the `block` in question to a bash command that does the equivalent. Said command would result in the icallsame shasum for the build instructions every time. This doesn't get around the issue that we can dynamically alter the build instructions based on the state of the system.
2. Steven has suggested that we move the `shasum` portion of the caching to take place before the build is execute, so that any side effects of the build are not present.
3. My concern is that we're even allowing DSL methods from within the `block` resource. Is that something that we want to do? How common is it?

stevendanna:
The Builder#shasum is based, in-part, on the build_commands in the
instance of Builder. This is problematic because the `block` DSL
method allows users to use other Builder DSL methods inside of it, all
of which mutate the build_command array when #build is called. Thus, a
given builder will return a different shasum before and after build is
called.

Caching shashum at the beginning of build fixes this problem, but has
a major drawback. Namely, two builders with the same shashum might
execute different commands.

This problem is partially mitigated by the fact that the builder
shashum is just a part of the software shasum, which will typically
include the actual text of the software definition as part of its
digest.

This will not catch software definitions where the build steps can
produce different based on the system state, such as this:

```
  block do
    Dir.glob("#{install_dir}/embedded/apath/bin/*").sort.each do |f|
      link bin, "#{install_dir}/embedded/bin/#{File.basename(f)}"
    end
  end
```

However, that will always be a problem since we allow pure Ruby in our
build steps.

Fixes #450

### DIFF
--- a/spec/unit/builder_spec.rb
+++ b/spec/unit/builder_spec.rb
@@ -180,5 +180,38 @@ module Omnibus
         subject.make(timeout: 3600)
       end
     end
+
+    describe "#shasum" do
+      let(:build_step) do
+        Proc.new {
+          block do
+            command("true")
+          end
+        }
+      end
+
+      let(:tmp_dir) { Dir.mktmpdir }
+      after { FileUtils.rmdir(tmp_dir) }
+
+      let(:software) do
+        double(Software,
+               name: 'chefdk',
+               install_dir: tmp_dir,
+               project_dir: tmp_dir,
+               overridden?: false)
+      end
+
+      let(:before_build_shasum) do
+        b = described_class.new(software)
+        b.evaluate(&build_step)
+        b.shasum
+      end
+
+      it "returns the same value when called before or after the build" do
+        subject.evaluate(&build_step)
+        subject.build
+        expect(subject.shasum).to eq(before_build_shasum)
+      end
+    end
   end
 end
--- a/lib/omnibus/builder.rb
+++ b/lib/omnibus/builder.rb
@@ -580,7 +580,8 @@ module Omnibus
     #
     def build
       log.info(log_key) { 'Starting build' }
-
+      shasum # ensure shashum is calculated before build since the build can alter the shasum
+      log.internal(log_key) { 'Cached builder checksum before build: #{shasum}'}
       if software.overridden?
         log.info(log_key) do
           "Version overridden from #{software.default_version} to "\
--- a/lib/omnibus/software.rb
+++ b/lib/omnibus/software.rb
@@ -519,7 +519,7 @@ module Omnibus
           }
         )
       end
- 
+
       env.merge(compiler_flags).
         merge(extra_linker_flags).
         # always want to favor pkg-config from embedded location to not hose


schisamo:
:+1:

spec/unit/builder_spec.rb
```diff
@@ -180,5 +180,38 @@ module Omnibus
...
+          end
+        }
+      end
+
+      let(:tmp_dir) { Dir.mktmpdir }
+      after { FileUtils.rmdir(tmp_dir) }
+
+      let(:software) do
+        double(Software,
+               name: 'chefdk',
```
marcparadise: Should this be fixed at 'chefdk'?

PR #458, opened 2015-04-01
`generate changelog` can now generate changelog between two refs

stevendanna:
This commit add support for the following:

- optional arguments for the starting and ending revisions to base the
  changelog off of.

- reading the start and end revisions from the appropriate version
  manifests

- reading the new version from the version manifest


stevendanna:
This should allow us to create a changelog given any two version manifests or any two git revisions.

cc @oferrigni

scotthain:
:+1: - not an expert on the functionality but the code looks good to me.

tylercloke:
LGTM :+1:

PR #460, opened 2015-04-09
ignore ctag files

curiositycasualty:
ignore ctag files for gems
ignore hidden ctag files
names for the file aren't connonical
because: grammar


scotthain:
:+1:

PR #466, opened 2015-04-24
Move to support 64-bit ruby on Windows.

ksubrama:


lib/omnibus/version.rb
```diff
@@ -15,5 +15,5 @@
 #
 
 module Omnibus
-  VERSION = '4.0.0'
+  VERSION = '4.0.1'
```
schisamo: Please do not bump versions like this in PRs. That is a release time task of which this change might be batched up with others.

omnibus.gemspec
```diff
@@ -23,9 +23,9 @@ Gem::Specification.new do |gem|
 
   gem.add_dependency 'chef-sugar',       '~> 3.0'
   gem.add_dependency 'cleanroom',        '~> 1.0'
-  gem.add_dependency 'mixlib-shellout',  '~> 1.4'
+  gem.add_dependency 'mixlib-shellout',  '~> 2.0'
   gem.add_dependency 'mixlib-versioning'
-  gem.add_dependency 'ohai',             '~> 7.2'
+  gem.add_dependency 'ohai',             '~> 8.2'
   gem.add_dependency 'ruby-progressbar', '~> 1.7'
```
schisamo: I'm a little worried about bump the major versions of both `mixlib-shellout` AND `ohai`. We'll need to do some pretty heavy testing that this doesn't break things on all the other platforms we build for.

omnibus.gemspec
```diff
@@ -23,9 +23,9 @@ Gem::Specification.new do |gem|
 
   gem.add_dependency 'chef-sugar',       '~> 3.0'
   gem.add_dependency 'cleanroom',        '~> 1.0'
-  gem.add_dependency 'mixlib-shellout',  '~> 1.4'
+  gem.add_dependency 'mixlib-shellout',  '~> 2.0'
   gem.add_dependency 'mixlib-versioning'
-  gem.add_dependency 'ohai',             '~> 7.2'
```
jdmundrawala: @schisamo do we use ruby 1.9 with ohai 4?

schisamo: We use Ruby 2.1.5 on all of our builders.

jdmundrawala:
I'm not sure this PR is necessary for us to build a 64 bit ruby package for chef

ksubrama:
I should withdraw the pull-request for now.  I was attempting to go down a "pure" 64-bit build process but I guess it's reasonable to have a 32-bit ruby embed a 64-bit ruby as part of the build.

PR #467, opened 2015-04-27
Correctly determine git repository when running in a subdirectory

stevendanna:
When running in a subdirectory of a git repository, .git won't exist.
Instead of looking for ./.git specifically, attempt to get the
revision and rescue if we can't.


stevendanna:
cc @chef/lob

schisamo:
:+1:

PR #476, opened 2015-05-08
Document allowable default_version settings for git sources.

yzl:
cc @chef/omnibus-maintainers

### DIFF
--- a/lib/omnibus/software.rb
+++ b/lib/omnibus/software.rb
@@ -256,13 +256,14 @@ module Omnibus
     expose :source
 
     #
-    # Set or retieve the {#default_version} of the software to build.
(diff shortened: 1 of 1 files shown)


lib/omnibus/software.rb
```diff
@@ -256,13 +256,14 @@ def source(val = NULL)
...
-    # Set or retieve the {#default_version} of the software to build.
+    # Set or retrieve the {#default_version} of the software to build.
     #
     # @example
     #   default_version '1.2.3'
     #
     # @param [String] val
-    #   the default version to set for the software
+    #   the default version to set for the software.
+    #   For a git source, the default version may be a sha or branch name.
```
stevendanna: It can also be a tag.

PR #477, opened 2015-05-11
Forcibly remove a non-empty project dir before cloning

stevendanna:
If multiple projects are using the same maching for building, they
might have two different software definitions that share the same name
but a different source. For example, the old omnibus-chef-server
repository uses a git-based source for bookhself. However, the new
chef-server repository uses a path-based source. If the new
chef-server project builds before the omnibus-chef-server project, the
build will fail with an error message from git complaining about a
non-empty directory.

NB: This clearly doesn't solve all the problems with having two
projects building on the same machine with different fetchers for a
given software. For example the two software definitions could both
use git but via different git upstreams.


schisamo:
:+1:

tylercloke:
:+1:

Issue #459, opened 2015-04-02
Preserve hard links in FileSyncer#sync?


jacobvosmaer:
We are seeing an interesting difference between the omnibus build product on our build boxes and on target systems after installing a DEB/RPM: hard links are not being preserved. When packaging Git with omnibus this seems to lead to a dramatic increase in disk use.

- FPM issue https://github.com/jordansissel/fpm/issues/365
- GitLab issue https://gitlab.com/gitlab-org/omnibus-gitlab/issues/274

On our builders, git sub-commands in libexec take up about 18MB:

```
$ ls -l /opt/gitlab/embedded/libexec/git-core/ | awk ' {total += $5 } END { print total/1024/1024 }'
259.11
$ du -sh /opt/gitlab/embedded/libexec/git-core 
18M	/opt/gitlab/embedded/libexec/git-core
```

But after installation, we are looking at 250+MB for the same directory:

```
$ du -sh /opt/gitlab/embedded/libexec/git-core/
260M	/opt/gitlab/embedded/libexec/git-core/
```

The problem is with hard links it seems, and I suspect that FileSyncer#sync is where the hard links get lost. `libexec/git-core` contains 100+ hard links to a single file. Considering how big omnibus packages can get to begin with, it might be nice if we could prevent exacerbating the situation by forgetting hard links.

jacobvosmaer:
This is how FPM ended up fixing the problem: https://github.com/jordansissel/fpm/commit/32c6be5488ed5a193498528c23620cb1c5aa5d31

curiositycasualty:
.jkeiser Walked me through how Chef handles this in the link/file provider. It's quite complex and ends up using a combination of a volume's serial number combined with the files "index" (https://github.com/chef/chef/blob/master/lib/chef/win32/file/info.rb#L83-L85) patched into a clone of the File class (file_class), both of which are obtained through the Win32 API.

jacobvosmaer:
On Unix-ish systems I would guess you could use `File.stat` and track the device/inode numbers, like the FPM patch does. I have no idea if `File.stat` contains useful/trustworthy information on Windows.

curiositycasualty:
The linux-only solution (for now) lives in https://github.com/chef/omnibus/pull/479
Preliminarily, Ruby's support for all the various Windows link-types looks extremely iffy; but we'll cross that bridge when we get to it.

curiositycasualty:
closing has fix merged

PR #479, opened 2015-05-12
accomidate hardlinks properly

curiositycasualty:
Address https://github.com/chef/omnibus/issues/459 by building/checking a list of hard link sources that have already been copied into the build directory ala https://github.com/jordansissel/fpm/blob/master/lib/fpm/util.rb#L160-L183

Discovered the hard way the `File.stat.ftype` != `File.ftype` ("File.stat.ftype" returns 'file' for both regular files and symlinks).

### DIFF
--- a/lib/omnibus/file_syncer.rb
+++ b/lib/omnibus/file_syncer.rb
@@ -101,16 +101,28 @@ module Omnibus
             FileUtils.ln_sf(target, "#{destination}/#{relative_path}")
           end
         when :file
-          # First attempt a regular copy. If we don't have write
-          # permission on the File, open will probably fail with
-          # EACCES (making it hard to sync files with permission
-          # r--r--r--). Rescue this error and use cp_r's
-          # :remove_destination option.
-          begin
-            FileUtils.cp(source_file, "#{destination}/#{relative_path}")
-          rescue Errno::EACCES
-            FileUtils.cp_r(source_file, "#{destination}/#{relative_path}",
-                           :remove_destination => true)
+          source_stat = File.stat(source_file)
+          # Detect 'files' which are hard links and use ln instead of cp to
+          # duplicate them, provided their source is in place already
+          if hardlink? source_stat
+            if existing = hardlink_sources[[source_stat.dev, source_stat.ino]]
+              FileUtils.ln(existing, "#{destination}/#{relative_path}")
+            else
+              FileUtils.cp(source_file, "#{destination}/#{relative_path}")
+              hardlink_sources.store([source_stat.dev, source_stat.ino], "#{destination}/#{relative_path}")
+            end
+          else
+            # First attempt a regular copy. If we don't have write
+            # permission on the File, open will probably fail with
+            # EACCES (making it hard to sync files with permission
+            # r--r--r--). Rescue this error and use cp_r's
+            # :remove_destination option.
+            begin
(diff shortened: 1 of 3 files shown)


scotthain:
Nice work. That's a gross bug. Is there any way to make a test for this?

schisamo:
Let's see if we can add some functional test coverage in https://github.com/chef/omnibus/blob/master/spec/functional/file_syncer_spec.rb

curiositycasualty:
@schisamo 
```
⮀ rspec ./spec/functional/file_syncer_spec.rb --color -fd
Run options:
  include {:focus=>true}
  exclude {:windows_only=>true}

All examples were filtered out; ignoring {:focus=>true}

Randomized with seed 7905

Omnibus::FileSyncer
  #sync
    when the destination is empty
      syncs the directories
    when target files are hard links
      and links to that instance thereafter
      copies the first instance
    with deeply nested paths and symlinks
      copies relative and absolute symlinks
    when the directory exists
      deletes existing files and folders
    when :exclude is given
      removes existing files and folders in destination
      does not copy files and folders that match the pattern
  #glob
    includes regular files
    ignores ..
    includes hidden files
    ignores .
    includes hidden folders

Finished in 0.15447 seconds (files took 0.47208 seconds to load)
12 examples, 0 failures
```

schisamo:
@curiositycasualty nice! looks like you are getting some failures on Ruby 2.x

stevendanna:
Modulo test failures, I'm :+1: on merging this.  Good sleuthing. 

While not necessary to merge, what do you think about some of the following refactor steps:

- The sync method is getting pretty long.  We might move some of the logic out of the case statement into functions like sync_file, sync_softlink, sync_hardlink. The case statement would then just dispatch out to those functions.

- In the case of a hardlink we haven't copied yet, I think that we want to use the same copy code that we use for regular files so that we don't run into the same permissions bug that we account for in that code.

- Perhaps we should have a helper function that encapsulates the file vs hardlink issue:

```ruby
def filetype(sourcefile)
  type = File.ftype(source_file).to_sym
  if type == :file && hardlink?( File.stat(source_file))
    :hardlink
  else
    type
  end
end
```

The beginning of the case statement could then just be `case filetype(sourcefile)` and you would have a new clause for hardlinks.

As an aside, bugs like this often makes me think: "We should just use rsync".

stevendanna:
Small nitpick, typo in your commit message :)

schisamo:
@stevendanna We purposely moved away from rsync to ensure Omnibus's file syncing worked across ALL platforms in a similar way. Rsync isn't always a good option on platforms like Windows.

curiositycasualty:
Huzzah! Tests are passing! :shipit: ?

scotthain:
I'm good on this as is, pending any large argument from @stevendanna - I think that's good feedback but should not be address during a bugfix?

stevendanna:
:+1: Nice work @curiositycasualty

maxlazio:
awesome work! Any chance of releasing a new version of omnibus with this included?

schisamo:
@maxlazio I'm hoping to get Omnibus 4.1 pushed to Rubygems.org next week sometime BUT you can just as easily update your `Gemfile` to pull in changes from the git repo directly:

```ruby
gem 'omnibus', github: 'chef/omnibus'
```

FWIW this is how Chef Software, Inc. handles things in all of our Omnibus projects.

PR #480, opened 2015-05-13
Avoid progress bar errors when downloading

danielsdeleo:
In Ci we randomly see errors from progress bar for trying to set the current progress higher than the total. My hunch is that this is the result of some computer arithmetic shenanigans and not a real problem with the downloaded file. Even if there's a real issue, we'll catch that in the next step when we verify the checksum of the downloaded thing, so there's no reason we need to put up with this error.

@chef/omnibus-maintainers

### DIFF
--- a/lib/omnibus/fetchers/net_fetcher.rb
+++ b/lib/omnibus/fetchers/net_fetcher.rb
@@ -169,8 +169,17 @@ module Omnibus
         format: '%e %B %p%% (%r KB/sec)',
         rate_scale: ->(rate) { rate / 1024 },
       )
-      options[:content_length_proc] = ->(total) { progress_bar.total = total }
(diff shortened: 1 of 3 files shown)


sdelano:
Only question left: do we still use the progress bar after this, and if not can we get rid of it?

Also, do you have example output of a failure?

sdelano:
FWIW, I'm :+1: on this.

danielsdeleo:
@sdelano the progress bar should still work. We're just using hacks with closures to make the max value available inside the proc that sets the current progress value so we can ensure we never exceed it.

schisamo:
:+1: :facepunch:

yzl:
:+1:

tyler-ball:
:+1:

sethvargo:
:+1: 

That's literally the worst API in the history of anything in Ruby core :frowning:

PR #487, opened 2015-06-04
Add disable_healthcheck option

mchlumsky:
This option disables the healthcheck.

The rationale is for the case where we'd like to build an omnibus that depends on OS-provided lib like OpenSSL When I patch OpenSSL (happens quite often) on the OS I won't need to rebuild the Omnibus. Currently, I can't even build the Omnibus because the healthchecker detects OpenSSL as an external dependency and exits.

Of course this seems to go against the whole point of an omnibus so I wont be too offended if you drop this pull request. ;)

### DIFF
--- a/lib/omnibus/config.rb
+++ b/lib/omnibus/config.rb
@@ -481,6 +481,11 @@ module Omnibus
       end
     end
 
+    # Disable the healthcheck
+    #
+    # @return [true, false]
+    default(:disable_healthcheck, false)
+
     # --------------------------------------------------
     # @!endgroup
     #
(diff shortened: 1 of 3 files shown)


scotthain:
@mchlumsky just as a side note - are you familiar with the the whitelist functionality? You can add specific files (system libraries if needed) to the software def to make it ignore them during the healthcheck. For example:

```whitelist_file 'libpcrecpp.so.0.0.0'```

Here's the code that allows you:
https://github.com/chef/omnibus/blob/4b019f031c32a0d92c389d3949103e643884d864/lib/omnibus/software.rb#L322

Whitelisting has worked well for me in the past, I don't know if we'd really want to encourage/allow disabling the full healthcheck, as you mentioned :)

curiositycasualty:
+1

mchlumsky:
@scotthain Thanks! It turns out whitelist_file does exactly what I need (just tested it) so I dont need this patch anymore.

scotthain:
@mchlumsky glad to hear it! I'm going to close this out and thanks for the contribution!

Issue #464, opened 2015-04-21
extra_package_file fails when passed a directory


tomashley:
If I pass a directory as an argument to extra_package_file in my project definition:

```
 [Software: version-manifest] I | Dirtied the cache
/opt/rubies/ruby-2.1.5/lib/ruby/2.1.0/fileutils.rb:1403:in `copy_stream': Is a directory - read (Errno::EISDIR)
        from /opt/rubies/ruby-2.1.5/lib/ruby/2.1.0/fileutils.rb:1403:in `block (2 levels) in copy_file'
        from /opt/rubies/ruby-2.1.5/lib/ruby/2.1.0/fileutils.rb:1402:in `open'
        from /opt/rubies/ruby-2.1.5/lib/ruby/2.1.0/fileutils.rb:1402:in `block in copy_file'
        from /opt/rubies/ruby-2.1.5/lib/ruby/2.1.0/fileutils.rb:1401:in `open'
        from /opt/rubies/ruby-2.1.5/lib/ruby/2.1.0/fileutils.rb:1401:in `copy_file'
        from /opt/rubies/ruby-2.1.5/lib/ruby/2.1.0/fileutils.rb:483:in `copy_file'
        from /opt/rubies/ruby-2.1.5/lib/ruby/2.1.0/fileutils.rb:400:in `block in cp'
        from /opt/rubies/ruby-2.1.5/lib/ruby/2.1.0/fileutils.rb:1579:in `block in fu_each_src_dest'
        from /opt/rubies/ruby-2.1.5/lib/ruby/2.1.0/fileutils.rb:1593:in `fu_each_src_dest0'
        from /opt/rubies/ruby-2.1.5/lib/ruby/2.1.0/fileutils.rb:1577:in `fu_each_src_dest'
        from /opt/rubies/ruby-2.1.5/lib/ruby/2.1.0/fileutils.rb:399:in `cp'
        from /home/vagrant/.gem/ruby/2.1.5/gems/omnibus-4.0.0/lib/omnibus/util.rb:167:in `copy_file'
        from /home/vagrant/.gem/ruby/2.1.5/gems/omnibus-4.0.0/lib/omnibus/packagers/rpm.rb:67:in `block (2 levels) in <class:RPM>'
        from /home/vagrant/.gem/ruby/2.1.5/gems/omnibus-4.0.0/lib/omnibus/packagers/rpm.rb:62:in `each'
[...]

PR #491, opened 2015-06-10
[bugfix] extra_pkg_file DSL allows directories

elafarge:
Since omnibus 4, the extra_package_file Project DSL wasn't allowing
directories anymore, which is kind of annoying when you're copying a dir
with a whole bunch of configuration files/examples files. Using
`FileUtils.cp_r` instead of `FileUtils.cp` in the `copy_file` util
method solves the issue.

It fixes https://github.com/chef/omnibus/issues/464

Obvious fix.

### DIFF
--- a/lib/omnibus/packagers/deb.rb
+++ b/lib/omnibus/packagers/deb.rb
@@ -35,13 +35,19 @@ module Omnibus
       # extra_package_file '/path/to/foo.txt' #=> /tmp/scratch/path/to/foo.txt
       project.extra_package_files.each do |file|
         parent      = File.dirname(file)
-        destination = File.join(staging_dir, parent)
 
-        create_directory(destination)
-        copy_file(file, destination)
+        if File.directory?(file)
+          destination = File.join(staging_dir, file)
+          create_directory(destination)
+          FileSyncer.sync(file, destination)
+        else
+          destination = File.join(staging_dir, parent)
+          create_directory(destination)
+          copy_file(file, destination)
+        end
       end
 
-      # Create the Debain file directory
+      # Create the Debian file directory
       create_directory(debian_dir)
     end
 
@@ -407,7 +413,7 @@ module Omnibus
           'armv6l'
         end
       when 'ppc64le'
-        # Debian prefers to use ppc64el for little endian architecture name 
+        # Debian prefers to use ppc64el for little endian architecture name
         # where as others like gnutools/rhel use ppc64le( note the last 2 chars)
         # see http://linux.debian.ports.powerpc.narkive.com/8eeWSBtZ/switching-ppc64el-port-name-to-ppc64le
         'ppc64el'  #dpkg --print-architecture = ppc64el
--- a/lib/omnibus/packagers/rpm.rb
+++ b/lib/omnibus/packagers/rpm.rb
@@ -61,10 +61,16 @@ module Omnibus
       # extra_package_file '/path/to/foo.txt' #=> /tmp/BUILD/path/to/foo.txt
       project.extra_package_files.each do |file|
         parent      = File.dirname(file)
-        destination = File.join("#{staging_dir}/BUILD", parent)
 
-        create_directory(destination)
-        copy_file(file, destination)
+        if File.directory?(file)
+          destination = File.join("#{staging_dir}/BUILD", file)
+          create_directory(destination)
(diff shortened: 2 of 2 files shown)


chef-supermarket:
Hi. I am an automated pull request bot named Curry. There are commits in this pull request whose authors are not yet authorized to contribute to Chef Software, Inc. projects or are using a non-GitHub verified email address. To become authorized to contribute, you will need to sign the Contributor License Agreement (CLA) as an individual or on behalf of your company. [You can read more on Chef's blog.](https://www.chef.io/blog/2014/06/23/changes-to-the-contributor-license-agreement-process)

## GitHub Users Who Are Not Authorized To Contribute

The following GitHub users do not appear to have signed a CLA:

* @elafarge

[Please sign the CLA here.](https://supermarket.chef.io/become-a-contributor)

chef-supermarket:
Hi. Your friendly Curry bot here. Just letting you know that all commit authors have become authorized to contribute. I have added the "Signed CLA" label to this issue so it can easily be found in the future.

elafarge:
Note : The PR has been updated to take into account the fact that we're using `FileUtils.cp_r` instead `FileUtils.cp`. Travis build is now passing.

yzl:
I'd prefer to respect the `omnibus` DSL convention of using `sync` for directories, but will defer to @sethvargo  and @schisamo.

sethvargo:
The notion for `copy_file` is that you are copying a single file. If you want directories, you should use `sync`!

scotthain:
I am +1 on using ```sync``` unless you have a use case it wouldn't work for... if you do we should probably fix sync.

elafarge:
It does make sense to use `sync` indeed. My original point was just to fix this issue: https://github.com/chef/omnibus/issues/464, which also happens to break backward compatibility with Omnibus 3. 

Would it be fine if I use `sync` instead of `copy_file` in the packagers, when processing extra files (here for instance https://github.com/chef/omnibus/blob/master/lib/omnibus/packagers/deb.rb#L41) ? I'll update and rebase the PR as soon as possible :)

yzl:
@elafarge `sync` [requires the source to be a directory](https://github.com/chef/omnibus/blob/master/lib/omnibus/file_syncer.rb#L66-70); the case where the extra file is a flat file would have to be handled separately.

elafarge:
@yzl I noticed that :) 

I updated the PR and launched a couple of builds, I'll test that I do have the expected behaviour asap.

sethvargo:
@yzl @schisamo looking at this a bit more, I think `extra_package_files` should just be an array of files. IMO, the user should use something like Dir.glob and array appending to populate the full list. Omnibus should just copy individual files (it makes the debug output much nicer).

yzl:
@elafarge Cool; I was writing in a hurry and ended up being overprecise. :-)

elafarge:
@sethvargo Sorry to reply so late. Your solution makes sense indeed, but the problem is that the `project` DSLs will be parsed at the beginning of the build. 

At that point, the files one wants to include in its package might not be there (because they are created/copied in that location during the `build` process) and therefore Dir.glob wouldn't find them. 

What do you think ? 

PS: this being said the current approach relying on `sync` is not working as expected as well but I could easily update the PR if you think it's worth it.

elafarge:
Hi @yzl @sethvargo :)

I updated & rebased the PR and ran a bunch of manual tests. The behavior is now the expected one.

The general idea is that it's better to use this DSL on real files when possible, I completely agree with that (especially since for most use cases it is just about shipping a couple of conf files to `/etc` and symlinks to `/usr/bin`. 

BUT for some directories, mostly directories with static content outside of the `install_dir` (like this one https://github.com/DataDog/dd-agent/tree/master/conf.d, containing static config files supposed to go in an `etc/dd-agent` folder), one might prefer to copy everything without having to look at it (because adding an `extra_package_file` every time a static file is added isn't fun and worse, one can easily forget to do so and have builds missing these new files).

Let me know if that use case makes sense to you. Also in case the PR won't be merged, this doc line will have to be updated https://github.com/chef/omnibus/blob/master/lib/omnibus/project.rb#L670 . (The `dir` mention will have to be removed).

Thanks for reviewing. 

Have a great day!

yzl:
@elafarge This looks reasonable.  Can you add a unit test for this case?

PR #493, opened 2015-06-16
Add warnings on empty globs

manderson26:
Copy/Move/Link silently do nothing if the glob provided doesn't match
any files. Since the intent for those commands almost certainly is to
actually *do* something I think they should warn if nothing actually
happens.

### DIFF
--- a/lib/omnibus/builder.rb
+++ b/lib/omnibus/builder.rb
@@ -488,10 +488,16 @@ module Omnibus
     # @return (see #command)
     #
     def copy(source, destination, options = {})
-      build_commands << BuildCommand.new("copy `#{source}' to `#{destination}'") do
+      command = "copy `#{source}' to `#{destination}'"
+      build_commands << BuildCommand.new(command) do
         Dir.chdir(software.project_dir) do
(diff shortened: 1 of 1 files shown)


jdmundrawala:
I think an empty glob should fail. Or least have an option for it to fail.

schisamo:
While I agree with @jdmundrawala (that an empty glob is probably an error) I am leery of changing the current behavior. The warning seems like a nice comprimise. What say you @chef/omnibus-maintainers?

sethvargo:
I think we should warn.

stevendanna:
I think we should warn here.  I would be :+1: on a global "fail on warnings" option that defaulted to off.

manderson26:
Sounds like generally +1.
Is everyone ok with merging this as is, where it warns but don't fail, and revisiting global fail-on-warn flag later?

schisamo:
@manderson26 I'm going to merge this now. Please [create another issue](https://github.com/chef/omnibus/issues/new) for tracking the global fail-on-warn flag feature.

PR #494, opened 2015-06-17
FileSyncer test for copying over read only file

seth:
In 4aa3116bb9, functionality was added to allow FileSyncer to copy over
a destination file with read-only mode. This patch adds an explicit test
case for this behavior.


stevendanna:
Good guy @seth, writing tests for my code :). :+1: LGTM

schisamo:
:+1: :ice_cream:

PR #495, opened 2015-06-17
Add ability to sign MSIs

jaym:
```ruby
package :msi do
  upgrade_code '{UID}'
  signing_identity 'THUMBPRINT'
end
```
cc @chef/omnibus-maintainers @chef/client-windows

### DIFF
--- a/lib/omnibus/exceptions.rb
+++ b/lib/omnibus/exceptions.rb
@@ -318,9 +318,9 @@ EOF
     end
   end
 
-  class FailedToTimestamp < Error
-    def initialize(msg)
-      super
+  class FailedToTimestampMSI < Error
+    def initialize
+      super("Failed to add timestamp to MSI.")
     end
   end
 end
--- a/lib/omnibus/exceptions.rb
+++ b/lib/omnibus/exceptions.rb
@@ -317,4 +317,10 @@ EOH
 EOF
     end
   end
+
+  class FailedToTimestamp < Error
+    def initialize(msg)
+      super
+    end
+  end
 end
--- a/lib/omnibus/packagers/msi.rb
+++ b/lib/omnibus/packagers/msi.rb
@@ -65,6 +65,9 @@ module Omnibus
 
         # Create the msi, ignoring the 204 return code from light.exe since it is
         # about some expected warnings
+
+        msi_file = windows_safe_path(Config.package_dir, package_name)
+
         light_command = <<-EOH.split.join(' ').squeeze(' ').strip
           light.exe
             -nologo
@@ -73,9 +76,13 @@ module Omnibus
             -cultures:en-us
             -loc "#{windows_safe_path(staging_dir, 'localization-en-us.wxl')}"
             project-files.wixobj source.wixobj
-            -out "#{windows_safe_path(Config.package_dir, package_name)}"
+            -out "#{msi_file}"
         EOH
         shellout!(light_command, returns: [0, 204])
+
+        if cert_name
+          sign_package(msi_file)
+        end
       end
     end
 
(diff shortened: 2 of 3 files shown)


danielsdeleo:
LGTM :+1:

lib/omnibus/exceptions.rb
```diff
@@ -317,4 +317,10 @@ def initialize(version)
 EOF
     end
   end
+
+  class FailedToTimestamp < Error
```
sethvargo: This seems like a pretty generic error, but it only appears to be used in the Windows timestamping. I think we do other kinds of timestampping. Would that error still apply and could be fixed in another PR? If this is a different kind of timestampping failure, maybe we could make the error more specific?

Also, we usually put the error text in the error itself (see above for examples). It looks like this is putting the error on the responsibility of the caller.

lib/omnibus/packagers/msi.rb
```diff
@@ -73,9 +76,13 @@ class Packager::MSI < Packager::Base
             -cultures:en-us
             -loc "#{windows_safe_path(staging_dir, 'localization-en-us.wxl')}"
             project-files.wixobj source.wixobj
-            -out "#{windows_safe_path(Config.package_dir, package_name)}"
+            -out "#{msi_file}"
         EOH
         shellout!(light_command, returns: [0, 204])
+
+        if cert_name
```
sethvargo: I think we should probably have an option for `sign` to align with the RPM UX. Then, if that option is true, we can validate that `cert_name` is present as a validation step.

lib/omnibus/packagers/msi.rb
```diff
@@ -178,6 +185,116 @@ def wix_candle_extension(extension)
...
+    #
+    # @param [Array] url
+    #   Urls of timestamp servers
+    #
+    # @return [Array]
+    #   A list of the timestamp servers to be used
+    #
+    def timestamp_servers(urls = NULL)
+      if null?(urls)
+        @timestamp_servers || ['http://timestamp.digicert.com',
```
sethvargo: There should be a constant

sethvargo: Can we use SSL?

lib/omnibus/packagers/msi.rb
```diff
@@ -343,5 +460,50 @@ def wix_candle_extensions
...
     end
+
+    #
+    # Takes a path to a msi and uses the set certificate store and 
+    # certificate name
+    #
+    def sign_package(msi_file)
+      raise MissingRequiredAttribute.new(self, :cert_store_name) if !cert_store_name
+      raise MissingRequiredAttribute.new(self, :cert_name) if !cert_name
+      shellout!("signtool.exe sign /v /s #{cert_store_name} /n #{cert_name} #{msi_file}")
```
sethvargo: Should we quote `msi_file`? What if the path contains spaces?

spec/unit/packagers/msi_spec.rb
```diff
@@ -263,5 +279,71 @@ module Omnibus
...
+
+
+          it 'raises an exception if there are no available timestamp servers' do
+            subject.timestamp_server('http://fooserver')
+            expect(subject).to receive(:try_timestamp).with(msi, 'http://fooserver').and_return(false)
+            expect {subject.sign_package(msi)}.to raise_error(FailedToTimestamp)
+          end
+        end
+      end
+
```
sethvargo: Extra newline

lib/omnibus/packagers/msi.rb
jdmundrawala: :+1: on the constant

cant use ssl. None of the timestamp servers provide that. I'm guessing it is ok because they are signing the hash that is provided with a timestamp, so signtool.exe has enough information to verify it

lib/omnibus/packagers/msi.rb
jdmundrawala: I'm not seeing this in RPM:
https://github.com/chef/omnibus/blob/master/lib/omnibus/packagers/rpm.rb#L307
https://github.com/chef/omnibus/blob/master/lib/omnibus/packagers/rpm.rb#L90-L96

sethvargo: It's https://github.com/chef/omnibus/blob/ce9639e12b50ce5b58e63a81aa0f09f29fc95315/lib/omnibus/config.rb#L239, but I think we are moving in a different direction so I guess this is fine. I just wonder if there is a use case where a person wants to specify the cert name, but skip signing (maybe for debugging).

jdmundrawala: It would be nice for everything to just work a dev box. So that means, I should be able to tell omnibus that it shouldn't sign anything without needing to comment stuff out in the project file.

sethvargo: Yea. Maybe that's another PR?

jdmundrawala: yea, it would need to touch all the packagers

lib/omnibus/packagers/msi.rb
```diff
@@ -178,6 +187,115 @@ def wix_candle_extension(extension)
...
+    # @example
+    #   cert_store_name "MY"
+    #
+    # @param [String] name
+    #   Name of the certificate store
+    #
+    # @return [String]
+    #   The name of the certificate store to be used
+    #
+    def cert_store_name(name = NULL)
```
schisamo: Can we please use a prefix of `singing_*` for these various signing config values. This would match the naming we use in other packagers and makes the purpose of these config values very obvious.

lib/omnibus/packagers/msi.rb
```diff
@@ -343,5 +461,50 @@ def wix_candle_extensions
...
       "#{arr.map {|e| "-ext '#{e}'"}.join(' ')}"
     end
+
+    #
+    # Takes a path to a msi and uses the set certificate store and 
+    # certificate name
+    #
+    def sign_package(msi_file)
+      raise MissingRequiredAttribute.new(self, :cert_store_name) if !cert_store_name
+      raise MissingRequiredAttribute.new(self, :cert_name) if !cert_name
```
schisamo: Small Ruby style/idiom thing: we usually use `unless foo` instead of `if !foo`.

lib/omnibus/packagers/msi.rb
schisamo: Should this be:

```ruby
if cert_store_name && cert_name
```

Would it make more sense to use a single config option:

```ruby
package :msi do
  upgrade_code "AB1D6FBD-F9DC-4395-BDAD-26C4541168E7"
  signing_certificate store: 'MY', name: 'MySigningCert'
end
```

jdmundrawala: How about a compromise between the two:
```ruby
signing_identity 'My Signing Cert'
```
with an optional `store` parameter. `signtool.exe` defaults this to `My`.  So if your cert was in some other store:
```ruby
signing_identity "My Signing Cert", store: "FooStore"
```
Unless there is a particular reason to use the fingerprint

sersut:
:+1: also looks like @jaym tested this is opscode-ci.

schisamo:
:+1:

PR #501, opened 2015-06-24
Use sudo rm -rf v. FileUtils for AIX #remove_directory

curiositycasualty:
Previously, the `/tmp/#{project}` folders weren't being removed leading to disks filling up.

### DIFF
--- a/lib/omnibus/packagers/bff.rb
+++ b/lib/omnibus/packagers/bff.rb
@@ -200,6 +200,24 @@ module Omnibus
       "#{safe_base_package_name}-#{project.build_version}-#{project.build_iteration}.#{safe_architecture}.bff"
     end
 
+    #
+    # Remove the directory at the given +path+ for AIX.
+    #
+    # @param [String, Array<String>] paths
+    #   the path or list of paths to join to delete
+    #
+    # @return [String]
+    #   the path to the removed directory
+    #
+    def remove_directory(*paths)
+      path = File.join(*paths)
+      log.debug(log_key) { "Remove directory `#{path}'" }
+      # we've already assumed sudo within
+      # #create_bff and /tmp is owned by
+      # root on AIX.
+      shellout!("sudo rm -rf \"#{path}\"")
+      path
+    end
 
     #
     # Return the BFF-ready base package name, converting any invalid characters to


schisamo:
:+1:

sethvargo:
Hrm. Is there a reason that the existing remove_directory with FileUtils.rm_rf doesn't work?

curiositycasualty:
@sethvargo /tmp is owned by root on AIX, FileUtils.rm_rf fails with a permissions error without sudo. This seemed like the most equitable solution + I was unsure how to elevate permissions for FileUtils.rm_rf in a more Ruby way.

sethvargo:
Hrm. It seems like we might want to choose a different temporary directory then, maybe something under `/tmp` that is owned by the omnibus Ruby? We did a great deal of work to remove the "unixisms" for omnibus as much as possible. Spawning a subshell to remove a file is a unixism and it would be great if we could work around it another way.

danielsdeleo:
I agree with @sethvargo here. According to the docs for `Dir.mktmpdir`, you can do something like `Dir.mktmpdir(nil, "/var/tmp") {|dir| dir is "/var/tmp/d..." }` to use an arbitrary directory instead of tmp. So it should work if we override `staging_dir` to do that.

curiositycasualty:
Guys.. I don't think it hurts to have a unixism that only appears in the AIX packaging portion of omnibus. This code will never run on a non-*nix platform (hopefully never on anything besides AIX..) @sethvargo @danielsdeleo

sethvargo:
@curiositycasualty it's not_just_ about the unixisms though - there is a larger issue which is that omnibus is writing to a directory that it cannot cleanup. Shelling out so we have sudo is a bandaid and introduces other problems (like what happens when the user doesn't have passwordless-sudo enabled?) - the build could hang waiting for a sudo password. That's just one example.

curiositycasualty:
A bandaid we've already used in #create_bff_file earlier, which makes the same assumptions of having passwordless-sudo setup for your build user. Changing the staging folder for the entirety of the omnibus codebase sounds like a large piece of work. Necessary work, but tangential to this PR.

danielsdeleo:
You don't have to change it for all of omnibus as far as I can tell. It appears the staging directory is only used by the packager class: https://github.com/chef/omnibus/blob/if/aix-remove-dirs/lib/omnibus/packagers/bff.rb#L30-L40 The BFF packager just inherits the default behavior from the base class: https://github.com/chef/omnibus/blob/if/aix-remove-dirs/lib/omnibus/packagers/base.rb#L175-L177 but you could easily override that method with a different behavior.

schisamo:
@curiositycasualty Can we take a stab at @danielsdeleo's suggested implementation approach of overriding `Omnibus::Packager::Base#staging_dir` in `Omnibus::Packager::BFF`? After reading the discussion this seems like the cleanest approach.

curiositycasualty:
Sorry folks, I think i was mistaken in my diagnosis. Below is how things look when they fail.
```
[11] pry(#<Omnibus::Packager::BFF>)> FileUtils.rm_rf(path, :verbose => true)
rm -rf /tmp/harmony20150630-15794314-zdqj2n
=> ["/tmp/harmony20150630-15794314-zdqj2n"]
[12] pry(#<Omnibus::Packager::BFF>)> .rm -rf #{path}
rm: 0653-609 Cannot remove /tmp/harmony20150630-15794314-zdqj2n/.info/.create_root.harmony.
The file access permissions do not allow the specified action.
rm: 0653-609 Cannot remove /tmp/harmony20150630-15794314-zdqj2n/.info/harmony.al.
The file access permissions do not allow the specified action.
<snip>
rm: 0653-611 Directory /tmp/harmony20150630-15794314-zdqj2n/usr is not empty.
rm: 0653-611 Directory /tmp/harmony20150630-15794314-zdqj2n is not empty.
Error: there was a problem executing system command: rm -rf /tmp/harmony20150630-15794314-zdqj2n
[13] pry(#<Omnibus::Packager::BFF>)> .ls -lad #{path}
drwx------    6 isa      staff           256 Jun 30 17:52 /tmp/harmony20150630-15794314-zdqj2n
[14] pry(#<Omnibus::Packager::BFF>)> .ls -la /tmp/harmony20150630-15794314-zdqj2n/opt/harmony/embedded/bin/entropy
-rwxr-xr-x    3 root     system         7135 Jun 30 17:52 /tmp/harmony20150630-15794314-zdqj2n/opt/harmony/embedded/bin/entropy
```
So the problem isn't that /tmp is owned by root, but more that a lot of the files remaining are owned by root (from the earlier `shellout!("sudo chown -R 0:0 #{staging_dir}/#{project.inst..`).

Is there a better way to escalate privileges to be able to remove these files than using `sudo` while shelling out? @danielsdeleo @sethvargo @schisamo

danielsdeleo:
My preferred option would be to not chown in the first place, but instead set the owner and group via packaging metadata. If that's not an option, then either Omnibus has to run as root and change `euid` and `egid` back and forth, or use your original solution.

stevendanna:
What's the state of this? Do we still need this? Has it been fixed some other way? If we still need this I think the current PR is probably the best near term solution.  We don't want omnibus running as root generally IMHO and the documentation I've read on `bff` so far has not made me hopeful that this is an option at the packaging level.

curiositycasualty:
@stevendanna We'll be in a better state to address this when more AIX CI work is done. I'll close this PR but leave the branch so I can investigate later.

PR #502, opened 2015-06-24
Clone git repositories with --recursive flag.

szechyjs:
If the repo contains any submodules this will initialize them after the clone completes.

### DIFF
--- a/lib/omnibus/fetchers/git_fetcher.rb
+++ b/lib/omnibus/fetchers/git_fetcher.rb
@@ -92,6 +92,15 @@ module Omnibus
     end
 
     #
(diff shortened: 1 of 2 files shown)


sethvargo:
I'm :+1: on this, but I wonder if there is a situation where a user would _not_ want submodules. Maybe it makes sense to have a `:submodules` option that defaults to false that can be specified in the software config? Like `source git: "...", submodules: true`. Thoughts?

szechyjs:
I like the idea of the :submodules option. I've pushed a change that includes the option.

yzl:
:+1:

schisamo:
:+1:

Issue #483, opened 2015-05-18
Add support for build dependencies


schisamo:
It would be nice for a software definition to declare a build dependency, e.g. something that is needed at build time but will not be shipped in the final package/artifact (they should also be excluded from any health checking). I envision this working via a new `build_dependency` software DSL function:

``` ruby
name "delivery-cli"
default_version "master"

source git: "https://github.com/chef/delivery-cli.git"

dependency "openssl"
build_dependency "rust"

build do
  env = with_standard_compiler_flags(with_embedded_path)

  command "cargo build -j #{workers} --release", env: env

  mkdir "#{install_dir}/bin"
  copy "#{project_dir}/target/release/delivery", "#{install_dir}/bin/delivery"
end
```

Under the covers the build dependency could be compiled and placed in a directory in `/var/cache/omnibus/*`. The environment returned from `with_standard_compiler_flags` would also be aware of this alternate location.

/cc @chef/omnibus-maintainers @chef/delivery

christophermaier:
:+1: This would be nice for 2 things for Delivery. 
* `delivery-cli` needs Rust to build, but not when deployed
* The web UI assets use NodeJS to build, but that's the only time we need it

danielsdeleo:
I don't see any reason why not.

szechyjs:
:+1: I've been looking forward to this feature for some time now.

lamont-granquist:
Could probably migrate a lot of the omnibus base installers that scotth has been working on into this feature if you did it so that these packages were installed somewhere that wasn't wiped between builds.

(the flip side of this argument would be to just get the rust compiler into those installers and take it out as a project dependency).

scotthain:
@danielsdeleo @lamont-granquist @szechyjs I put up a PR that has code that may solve this issue - please take a look and try it out if you get a chance.

lamont-granquist:
embedded libtool broke chef-dk stuff for me:

https://github.com/chef/chef-dk/issues/682

we need a way of having build-deps that we do not ship in the package one way or the other.

scotthain:
I did a 2nd POC that shows one approach (and actually works this time) https://github.com/chef/omnibus/tree/shain/build_dependencies

brodock:
Any progress into that issue?

szechyjs:
@scotthain any plans on submitting a pull request for your 2nd POC?

schisamo:
[...]

PR #504, opened 2015-06-26
First take at a build dependency feature

scotthain:
This is mostly a POC to see if it would solve the issue described in https://github.com/chef/omnibus/issues/483

Based on the requirements in the issue, I think it will work. Any feedback is appreciated.

To use this, simply add a ```build_dependency 'my_def'``` in your software definition and it will build the dependency along with the rest of your things and add it to the path. At that point it will be available in the ```with_standard_compiler_flags``` created environment and you can use it in a ```command``` dsl method.

### DIFF
--- a/lib/omnibus/software.rb
+++ b/lib/omnibus/software.rb
@@ -29,7 +29,7 @@ module Omnibus
       #
       # @return [Software]
       #
-      def load(project, name, manifest)
+      def load(project, name, manifest, build_dep = false)
         loaded_softwares[name] ||= begin
           filepath = Omnibus.software_path(name)
 
@@ -41,9 +41,10 @@ module Omnibus
             end
           end
 
-          instance = new(project, filepath, manifest)
+          instance = new(project, filepath, manifest, build_dep)
           instance.evaluate_file(filepath)
           instance.load_dependencies
+          instance.load_build_dependencies
 
           # Add the loaded component to the library
           project.library.component_added(instance)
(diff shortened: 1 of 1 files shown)


lib/omnibus/software.rb
```diff
@@ -696,6 +750,11 @@ def overridden?
       @overrides.key?(:version) && (@overrides[:version] != @version)
     end
 
+    def build_dep
+      @build_dep
+    end
+    expose :build_dep
```
schisamo: We should name this `build_dependency` so it matches with the naming idioms of `dependency`.

scotthain: I couldn't name it that because there already exists a build_dependency method that is exposed and does something different. I'm totally down with finding a better way of doing that, but couldn't figure it out... such metaprogramming...

schisamo: @scotthain Can you link to the existing `build_dependency` method definition? I'm having a hard time finding it.

scotthain: Line 214 just above

sethvargo: I don't think that other method actually does anything

scotthain: Entirely possible @sethvargo :) I will debug through and see if it gets triggered.

scotthain:
Something is goofed on this branch, the code is not all here... I will debug and send a new PR in. Sorry about the confusion!

PR #505, opened 2015-06-29
Fix reading of build_version from manifest files.

szechyjs:
When generating a changelog like so the build_version is nil.

`omnibus changelog generate --starting_manifest start.json --ending_manifest current.json --minor false`

I tracked this down to the parser not populating the fields.

/cc @schisamo @stevendanna

### DIFF
--- a/lib/omnibus/manifest.rb
+++ b/lib/omnibus/manifest.rb
@@ -92,7 +92,7 @@ module Omnibus
     end
 
     def self.from_hash_v1(manifest_data)
-      m = Omnibus::Manifest.new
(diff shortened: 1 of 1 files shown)


danielsdeleo:
Looks pretty straightforward. Could use a regression test.

schisamo:
:+1:, but I'm with @danielsdeleo, would like to see a test added.

schisamo:
@szechyjs I pulled your fix into https://github.com/chef/omnibus/pull/508. I also added test coverage.

PR #508, opened 2015-07-02
Create an Artifactory build for published packages

schisamo:
An Artifactory build gives us a top level object that groups a set of packages together.

* This build object is the object we use to drive promotions and cleanup tasks.
* This build object is where we can attach build-wide metadata, things like the build’s version-manifest.json.

See the following for more details on what 'build' means in the Artifactory world:
https://www.jfrog.com/confluence/display/RTF/Artifactory+REST+API#ArtifactoryRESTAPI-BUILDS
https://www.jfrog.com/confluence/display/RTF/Build+Integration
https://github.com/JFrogDev/build-info

Some shots of what is produced:

artifactory_chef_s_preprod_artifactory_instance____build_browser

artifactory_chef_s_preprod_artifactory_instance____build_browser

artifactory_chef_s_preprod_artifactory_instance____build_browser

artifactory_chef_s_preprod_artifactory_instance____repository_browser

/cc @chef/engineering-services @chef/omnibus-maintainers

### DIFF
--- a/features/step_definitions/generator_steps.rb
+++ b/features/step_definitions/generator_steps.rb
@@ -1,7 +1,7 @@
 require 'aruba/api'
 
 Given(/^I have an omnibus project named "(.+)"$/) do |name|
-  create_dir(name)
+  create_directory(name)
   cd(name)
 
   write_file("config/projects/#{name}.rb", <<-EOH.gsub(/^ {4}/, ''))
--- a/features/step_definitions/generator_steps.rb
+++ b/features/step_definitions/generator_steps.rb
@@ -27,3 +27,15 @@ Given(/^I have an omnibus project named "(.+)"$/) do |name|
     package_tmp   './local/omnibus/pkg-tmp'
   EOH
 end
+
+Given(/^I have a platform mappings file named "(.+)"$/) do |name|
+  write_file(name, <<-EOH.gsub(/^ {4}/, ''))
+    {
+      "ubuntu-10.04": [
+        "ubuntu-10.04",
+        "ubuntu-12.04",
+        "ubuntu-14.04"
+      ]
+    }
+  EOH
+end
--- a/lib/omnibus/cli.rb
+++ b/lib/omnibus/cli.rb
@@ -85,7 +85,7 @@ module Omnibus
       if @options[:output_manifest]
         FileUtils.mkdir_p('pkg')
         File.open(::File.join('pkg', 'version-manifest.json'), 'w') do |f|
-          f.write(JSON.pretty_generate(project.built_manifest.to_hash))
+          f.write(project.built_manifest.to_json)
         end
       end
     end
--- a/lib/omnibus/cli/publish.rb
+++ b/lib/omnibus/cli/publish.rb
@@ -18,19 +18,23 @@ module Omnibus
   class Command::Publish < Command::Base
     namespace :publish
 
-    # These options are useful for publish packages that were built for a
-    # paticluar platform/version and tested on another platform/version.
+    # This option is useful for publish packages that were built for a
+    # particular platform/version but tested on other platform/versions.
     #
     # For example, one might build on Ubuntu 10.04 and test/publish on
-    # Ubuntu 11.04, 12.04 and Debian 7.
+    # Ubuntu 10.04, 12.04, and 14.04.
     #
-    # If these options are used with the glob pattern support all packages
-    # will be published to the same platform/version.
-    class_option :platform,
-      desc: 'The platform to publish for',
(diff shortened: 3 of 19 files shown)


Gemfile
```diff
@@ -1,6 +1,8 @@
 source 'https://rubygems.org'
 gemspec
 
+gem 'artifactory', git: 'https://github.com/chef/artifactory-client.git', branch: 'schisamo/build-support'
```
danielsdeleo: I assume this will be cleaned up

lib/omnibus/cli.rb
```diff
@@ -85,7 +85,7 @@ def build(name)
       if @options[:output_manifest]
         FileUtils.mkdir_p('pkg')
         File.open(::File.join('pkg', 'version-manifest.json'), 'w') do |f|
-          f.write(JSON.pretty_generate(project.built_manifest.to_hash))
+          f.write(project.built_manifest.to_json)
```
danielsdeleo: will this unpretty the JSON?

danielsdeleo: Just scrolled down and see the answer is no. NM

Gemfile
schisamo: @danielsdeleo Yep. Commit message for that change says:

```
[WIP] Point at Artifactory client branch with build support

This will go away when chef/artifactory-client#53 is merged.
```

schisamo:
One more round of review please @chef/omnibus-maintainers and @chef/engineering-services

features/step_definitions/generator_steps.rb
```diff
@@ -1,7 +1,7 @@
 require 'aruba/api'
 
 Given(/^I have an omnibus project named "(.+)"$/) do |name|
-  create_dir(name)
+  create_directory(name)
```
sethvargo: Ugh - I just had to do this in ChefSpec. It was sad :frowning:

lib/omnibus/cli/publish.rb
```diff
@@ -68,8 +76,13 @@ def artifactory(repository, pattern)
     # @return [void]
     #
     def publish(klass, pattern, options)
+
```
sethvargo: Newline

lib/omnibus/metadata.rb
```diff
@@ -212,7 +212,7 @@ def truncate_platform_version(platform_version, platform)
     #
     def initialize(package, data = {})
       @package = package
-      @data    = data.dup.freeze
```
sethvargo: Hrm. I'm hesitant to unfreeze this. Can we keep it frozen and then `.dup` the `to_hash`? That will unfreeze it

lib/omnibus/publishers/artifactory_publisher.rb
```diff
@@ -56,6 +59,8 @@ def publish(&block)
         # If a block was given, "yield" the package to the caller
         block.call(package) if block
       end
+
+      build_for(packages).save unless packages.empty?
```
sethvargo: Should we log a warning/message when there are no packages?

lib/omnibus/metadata.rb
schisamo: oooh...awesome tip/advice, was having trouble finding away around the freeze.

lib/omnibus/publisher.rb
```diff
@@ -61,7 +66,50 @@ def initialize(pattern, options = {})
...
-      @packages ||= FileSyncer.glob(@pattern).map { |path| Package.new(path) }
+      @packages ||= begin
+        publish_packages = Array.new
+        build_packages   = FileSyncer.glob(@pattern).map { |path| Package.new(path) }
+
+        if @options[:platform_mappings]
+          # the platform map is a simple hash with publish to build platform mappings
+          @options[:platform_mappings].each_pair do |build_platform, publish_platforms|
+            # Splits `ubuntu-12.04` into `ubuntu` and `12.04`
+            build_platform, build_platform_version = build_platform.split('-')
```
danielsdeleo: Is there any time there would be a hyphen in either the platform or the version? Of course, it will be hard to handle both cases, but if you're only concerned about one of them, you could use `String#partition` or `String#rpartition` instead. Or you could check that the number of elements returned from split is what you expected.

danielsdeleo:
:+1: minor comment aside

lib/omnibus/publisher.rb
schisamo: Awesome tip @danielsdeleo, I think `String#rpartition` is just what is needed. If there is an extra hyphen it will most likely appear in the platform.

PR #513, opened 2015-07-16
Add libkvm and libprocstat to freebsd whitelist

jaym:
Ruby 2.2 wants these things on freebsd


jaym:
cc @lamont-granquist @thommay

lamont-granquist:
Yeah that is most likely fine :+1:

PR #515, opened 2015-07-23
Add support for building on Cisco's Nexus platform.

mattray:
Nexus runs on top of Wind River Linux 5, which uses RPMs but is not Red Hat-based.
Ohai patches for WRL5 have been submitted and we (@edolnx and I) have unlocked
the ohai version to pick up the master branch.

### DIFF
--- a/lib/omnibus/health_check.rb
+++ b/lib/omnibus/health_check.rb
@@ -425,7 +425,7 @@ module Omnibus
       whitelist_libs = case Ohai['platform']
                        when 'arch'
                          ARCH_WHITELIST_LIBS
-                       when 'wrlinux'
+                       when 'nexus'
                          WRLINUX_WHITELIST_LIBS
                        when 'mac_os_x'
                          MAC_WHITELIST_LIBS
--- a/lib/omnibus/health_check.rb
+++ b/lib/omnibus/health_check.rb
@@ -52,6 +52,17 @@ module Omnibus
     ].freeze
 
     WRLINUX_WHITELIST_LIBS = [
+      /libc\.so/,
+      /libcrypt\.so/,
+      /libdb-5\.3\.so/,
+      /libdl\.so/,
+      /libffi\.so/,
+      /libgdbm\.so/,
+      /libm\.so/,
+      /libnsl\.so/,
+      /libpthread\.so/,
+      /librt\.so/,
+      /libutil\.so/,
       /libffi\.so/,
     ].freeze
 
--- a/lib/omnibus/health_check.rb
+++ b/lib/omnibus/health_check.rb
@@ -51,6 +51,10 @@ module Omnibus
       /libutil\.so/,
     ].freeze
 
+    WRLINUX_WHITELIST_LIBS = [
+      /libffi\.so/,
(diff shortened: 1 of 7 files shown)


mattray:
Test is failing because it's running on Ruby 1.9 and ohai needs 2.0 or later. Open another PR to update that?

omnibus.gemspec
```diff
@@ -23,9 +23,9 @@ Gem::Specification.new do |gem|
 
   gem.add_dependency 'chef-sugar',       '~> 3.0'
   gem.add_dependency 'cleanroom',        '~> 1.0'
-  gem.add_dependency 'mixlib-shellout',  '~> 1.4'
+  gem.add_dependency 'mixlib-shellout'
   gem.add_dependency 'mixlib-versioning'
-  gem.add_dependency 'ohai',             '~> 7.2'
+  gem.add_dependency 'ohai'
```
schisamo: We should not remove the pessimistic pinning for mixlib-shellout OR ohai.

lib/omnibus/health_check.rb
```diff
@@ -35,6 +35,7 @@ class HealthCheck
       /libutil\.so/,
       /linux-vdso.+/,
       /linux-gate\.so/,
+      /libffi\.so/,
```
schisamo: Why are we whitelisting this? libFFI is not a system library.

omnibus.gemspec
edolnx: This prevents us from properly detecting any platforms added since ohai
7.2, and we have several projects which require new platforms to be
added and supported. Should we be backporting all new platform support
to 7.2 instead?

lib/omnibus/health_check.rb
edolnx: Building on new platforms (specifically for Cisco) the healthcheck for
the chef suite was failing because libffi was not on the allowed list.
This corrected that.

schisamo: This points to a linking bug in the underlying software definitions. Something isn't linking against the libffi we build but instead linking against the system libffi. The health check if failing just as it should in this situation.

omnibus.gemspec
danielsdeleo: We could upgrade to `~> 8.0` as long as that works. We just don't want to float on the latest, lest 9.0 break us.

schisamo: I wasn't questioning using a later version of mixlib-shellout or ohai but we should still pessimistically lock to the newer Ohai version.

@chef/omnibus-maintainers what are your thoughts on bumping the Ohai version and removing Ruby 1.9 support?

danielsdeleo: I like it in principle, and if Ohai 8.x works then I'll like it in practice.

lamont-granquist: `ohai ~> 8.0` seems like the right thing to do.  ruby 1.9 is EOL and RIP.

lib/omnibus/health_check.rb
lamont-granquist: Yep this is the wrong fix.

PR #516, opened 2015-07-29
Properly render ERB templates that do not have variables

ryancragun:
Fixes an error when rendering an ERB template that does not contain variables

```
Builder: chef-marketplace-cookbooks] I | Build chef-marketplace-cookbooks: 0.7814s
/home/vagrant/.gem/ruby/2.1.5/bundler/gems/omnibus-e28303ad28a6/lib/omnibus/templating.rb:56:in `new': wrong number of arguments (0 for 1+) (ArgumentError)
        from /home/vagrant/.gem/ruby/2.1.5/bundler/gems/omnibus-e28303ad28a6/lib/omnibus/templating.rb:56:in `render_template'
```

### DIFF
--- a/lib/omnibus/exceptions.rb
+++ b/lib/omnibus/exceptions.rb
@@ -87,7 +87,7 @@ EOH
 
     def to_s
       <<-EOH
-Attempting to evaluate the template `#{@name}', but it was not found at any of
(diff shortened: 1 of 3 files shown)


schisamo:
:+1: 

/cc @chef/omnibus-maintainers

sethvargo:
:+1:

PR #517, opened 2015-07-31
Added documentation about whitelisting libraries

scotthain:
@chef/omnibus-maintainers

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,6 +269,20 @@ omnibus manifest PROJECT -l warn
 This will output a JSON-formatted manifest containing the resolved
 version of every software definition.
 
+Whitelisting Libraries
+----------------------
+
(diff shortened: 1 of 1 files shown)


danielsdeleo:
Can you add some guidelines to help people understand the health check and when they should or should not whitelist? E.g., "you should only add libraries to the whitelist that are guaranteed to be on the system you install to, if a library comes from a non-default package you should instead build it into the package"

scotthain:
@danielsdeleo great idea, I borrowed exactly what you said there because I couldn't think of a better way to explain it :)

yzl:
lgtm

danielsdeleo:
:+1:

schisamo:
:+1:

PR #518, opened 2015-08-05
Add a "windows_arch" omnibus option to choose 32/64 bit builds.

ksubrama:


ksubrama:
:construction: work in progress

ksubrama:
:+1:
I created the pull request but this is @chefsalim's work.  /cc @schisamo @adamedx

lib/omnibus/packagers/msi.rb
```diff
@@ -407,6 +410,17 @@ def wix_candle_extensions
...
 
     #
+    # Returns the options to use for candle
+    #
+    # @return [Array]
+    #   the extensions that will be loaded for candle
+    #
+    def wix_candle_flags
+      # we support x86 or x64.  No Itanium support (ia64).
+      @wix_candle_flags ||= "-arch " + (Config.windows_arch == :x86 ? "x86" : "x64")
```
ksubrama: You can use "to_s" to convert the symbol to a string, but the fact that the symbol names correspond to the candle flags is a coincidence...  so *shrug*.  You can leave it as is if you feel it's cleaner.

lib/omnibus/config.rb
```diff
@@ -443,6 +443,16 @@ def reset!
...
 
+    # Architecture to target when building on windows.  This option
+    # should affect the bit-ness of Ruby and DevKit used, the platform of
+    # any MSIs generated and package dlls being downloaded.
+    #
+    # See the windows_x86? and windows_x64? software definition dsl
+    # methods.
+    #
+    # @return [:x86, :x64]
+    default(:windows_arch, :x86)
```
schisamo: I think we want to the default should be determined by whatever Ohai returns:

```ruby
default(:windows_arch) do
  if Ohai['kernel']['machine'] == 'x86_64'
    :x64
  else
    :x86
  end
end
```

lib/omnibus/config.rb
```diff
@@ -443,6 +443,16 @@ def reset!
     # @return [String, nil]
     default(:solaris_linker_mapfile, "files/mapfiles/solaris")
 
+    # Architecture to target when building on windows.  This option
+    # should affect the bit-ness of Ruby and DevKit used, the platform of
+    # any MSIs generated and package dlls being downloaded.
+    #
+    # See the windows_x86? and windows_x64? software definition dsl
```
schisamo: Looks like you named your method `windows_arch_i386?`.

lib/omnibus/config.rb
ksubrama: That would break the builds we have right now.  All our machines are 64-bit. And we still want to only be generating 32-bit builds for the foreseeable future until we get the matrix right and have other builders added.

schisamo: @ksubrama Then we can explicitely set that in omnibus-chef's [Omnibus config file](https://github.com/chef/omnibus-chef/blob/master/omnibus.rb):

```ruby
windows_arch :x86
```

My point is the default value should be programmatically set via Ohai data.

ksubrama: Ah fair enough.  And I guess because omnibus-chef has a Gemfile.lock, Manhattan won't pick up this change until we update it which we can do so in the same pull-request during which we add that option to the config file.

chefsalim:
This flag is for the build target, not the overall architecture (where Ohai setting the default value absolutely would be the right thing to do!).  In this case, since the _current_ builds are x86, I'm aiming to have this check in to preserve the status quo with a x86 build target as we have it today (until we are ready to do the dual builds with the rest of the pipeline). Perhaps I should just rename the flag to "windows_build_target" be clear on this.

chefsalim:
Never mind, I missed the part about override in omnibus.rb :)

jaym:
@schisamo are we ok with this breaking backwards compatibility?

ksubrama:
@jaym https://github.com/chef/omnibus-chef/pull/458 should handle the migration path for us.

jaym:
right, but that's just for us. As far as I understand, this will break other omnibus users. Perhaps a major version bump is required here?

ksubrama:
Ah yes...  maybe I should just bump version.rb to 5.0.0, in case someone's on the bleeding edge.

PR #475, opened 2015-05-07
Add Appveyor for Windows testing

schisamo:
/cc @chef/omnibus-maintainers

PR #519, opened 2015-08-14
Change git fetcher to correct fetch when repo name is the same but branch/version is different.

ksubrama:


ksubrama:
/cc @schisamo @sersut @danielsdeleo - this is the fix to the issue we ran into on manhattan when we changed berkshelf's source/version to Dan's.  Can you /cc the omnibus-maintainers as well?

Most of the test fixes are not related to the change.  It's just to get stuff to pass on windows because expand_path does things on windows...

danielsdeleo:
:+1: LGTM. We should get this going on appveyor so we can keep it passing on windows

schisamo:
:+1: I agree with @danielsdeleo, any chance we can get Appveyor wired into this repo?

stevendanna:
In case other reviewers were wondering, the change that fixes the bug described in the PR description is here:

https://github.com/chef/omnibus/commit/84c9dc454f6746935528d4eb991f003c9f988e68#diff-668e7c16f991ed53a3601327c77077ddR154

lib/omnibus/fetchers/git_fetcher.rb
```diff
@@ -193,6 +192,13 @@ def git(command)
 
     def self.resolve_version(ref, source)
       if sha_hash?(ref)
+        # A git server negotiates in terms of refs during the info-refs phase of a fetch.
+        # During upload-pack, the client is not allowed to specify any sha1s in the "wants"
+        # unless the server has publicized them during info-refs.  Hence, the server is allowed
+        # to drop requests to fetch particular sha1s, even if it is an otherwise reachable
+        # commit object.  Only when the service is specifically configured with
+        # uploadpack.allowReachableSHA1InWant is there any guarantee that it consider "naked" wants.
+        log.warn(log_key) { 'A sha1 hash is not a ref name and there is no guarantee that you will be fetching the commits you need.'}
```
stevendanna: Based on how we resolve the manifest currently, isn't this going to warn on pretty much every git-based software definition we have?

lib/omnibus/fetcher.rb
```diff
@@ -44,6 +44,12 @@ class Fetcher
     attr_reader :resolved_version
 
     #
+    # The upstream version as described before resolution.  This
+    # will usually be the same as +resolved_version+ but may refer to
+    # a remote ref name or tag for a source such as git.
+    attr_reader :described_version
```
sethvargo: Need an `@return` here for the content type. The general style we use is also to end with an extra comment newline before the attr_reader. See above for examples please.

lib/omnibus/fetchers/git_fetcher.rb
```diff
@@ -151,9 +150,9 @@ def git_checkout(ref=resolved_version)
     #
     # @return [void]
     #
-    def git_fetch(ref=resolved_version)
-      git("fetch --all")
-      git("reset --hard #{ref}")
+    def git_fetch
+      git("fetch #{source_url} #{described_version}#{' --recurse-submodules=on-demand' if clone_submodules?}")
```
sethvargo: Let's try not to be too tricky here. This violates the 80-character line-length, but it's also difficult to understand what is going on. Even though it's longer, I think the following is much clearer and less likely to cause confusion in the future:

```ruby
fetch_cmd = "fetch #{source_url} #{described_version}"
fetch_cmd << " --recurse-submodules=on-demand" if clone_submodules?
git(fetch_cmd)
```

What do you think?

lib/omnibus/fetchers/git_fetcher.rb
```diff
@@ -205,7 +211,7 @@ def self.resolve_version(ref, source)
     # @return [true, false]
     #
     def self.sha_hash?(rev)
-      rev =~ /^[0-9a-f]{4,40}$/
+      rev =~ /^[0-9a-fA-F]{4,40}$/
```
sethvargo: Let's just add `/i` to the end to make it case insensitve; that's more efficient anyway since it's a smaller scan set and a numeric comparison operation.

lib/omnibus/fetchers/git_fetcher.rb
sethvargo: - 80 characters on comment
- 80 characters on warning
- double spaces after period on comment do not match the rest of the project style
- I agree with @stevendanna - what is the "right" way to specify a version to resolve. Most of this is taken straight from Rubygem's Bundler

lib/omnibus/packagers/bff.rb
```diff
@@ -61,7 +61,7 @@ def package_name
     # @return [String]
     #
     def scripts_install_dir
-      File.expand_path("#{project.install_dir}/embedded/share/installp")
+      File.expand_path('embedded/share/installp', project.install_dir)
```
sethvargo: This actually isn't the same thing:

```irb
[1] pry(main)> File.expand_path("foo/bar/baz", "a/b/c")
=> "/Users/sethvargo/Development/vault-rails/a/b/c/foo/bar/baz"
[2] pry(main)> File.expand_path("a/b/c/foo/bar/baz")
=> "/Users/sethvargo/Development/vault-rails/a/b/c/foo/bar/baz"
[3] pry(main)> File.expand_path("/foo/bar/baz", "a/b/c")
=> "/foo/bar/baz"
[4] pry(main)> File.expand_path("foo/bar/baz", "/a/b/c")
=> "/a/b/c/foo/bar/baz"
```

sethvargo: We can use `File.join`, but expanding via relative paths was basically ripped out of omnibus because of the unexpected behavior with relative vs absolute paths vs the cwd.

lib/omnibus/packagers/bff.rb
```diff
@@ -173,16 +173,16 @@ def create_bff_file
...
       # project owned by the build user (which is incorrect)
-      shellout!("sudo chown -R 0:0 #{staging_dir}/#{project.install_dir.match(/^\/?(\w+)/)}")
+      shellout!("sudo chown -R 0:0 #{File.join(staging_dir, project.install_dir)}")
       log.info(log_key) { "Creating .bff file" }
 
       # Since we want the owner to be root, we need to sudo the mkinstallp
       # command, otherwise it will not have access to the previously chowned
       # directory.
-      shellout!("sudo /usr/sbin/mkinstallp -d #{staging_dir} -T #{staging_dir}/gen.template")
+      shellout!("sudo /usr/sbin/mkinstallp -d #{staging_dir} -T #{File.join(staging_dir, 'gen.template')}")
```
sethvargo: I don't really understand how only changing the staging_dir to a File.join fixes anything? You still have a forward slash in the mkinstallp path.

spec/functional/builder_spec.rb
```diff
@@ -9,8 +9,10 @@ module Omnibus
     # for testing methods like +ruby+ and +rake+ without the need to compile
     # a real Ruby just for functional tests.
     #
+    # Haha, this will totally work on windows..  (no it won't).
```
sethvargo: These types of comments are generally not helpful. Funny, but not helpful. Imagine a future engineer reading this. **Why** doesn't it work on Windows? Why the sarcasm? Can I refactor this?

sethvargo: So, to ask the obvious - why are these tests supported on Windows?

spec/support/file_helpers.rb
```diff
@@ -39,8 +39,12 @@ def create_file(*paths, &block)
         path
       end
 
-      def create_link(a, b)
-        FileUtils.ln_s(a, b)
+      def create_link(src, dest)
```
sethvargo: A comment here would be nice explaining things

spec/unit/package_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 module Omnibus
   describe Package do
-    let(:path) { '/path/to/files/thing.deb' }
+    let(:path) { "#{windows? ? 'C:' : ''}/path/to/files/thing.deb" }
```
sethvargo: This seems hacky, and having a ternary operator in an interpolation operation is very difficult to follow.

sethvargo:
Hi @ksubrama 

I did a line-by-line review, but I have some high-level questions/comments:

1. How do you recommend checking out a specific SHA? I think that warning is going to be super verbose and cause a lot of trouble for the eng team looking at builds.
2. Please refrain from making unnecessary changes in the PR. There are a lot of instances of changing quote styles that are unrelated to this PR. I understand that you probably changed them as you went along (I've been there too), but it makes the review exponentially more difficult in trying to read what was changed. There are entire files that could have been excluded from the diff because it was just a simple quote change. Those are better left to a "style-only" PR or at least isolated to their own commit. This is great work, but I would by lying if I said it wasn't fairly painful to review :smile: 
3. Please take a look at the existing project style, especially regarding comments. We want to keep the project as maintainable and consistent as possible.

Please let me know if you have any questions.

lib/omnibus/fetchers/git_fetcher.rb
ksubrama: The manifest holds both described_version and resolved_version.  We just need to ensure that described_version always points to a git tag or branch or a valid ref name otherwise.  That's why I ensure that in the constructor, I extract both into the fetcher's attributes and use the described_version to fetch from the git source and the resolved version during git checkout.

lib/omnibus/fetchers/git_fetcher.rb
ksubrama: Agreed :)  I still speak ruby with a LISP and am trying to get better at it.

ksubrama:
That's a very thorough code review - thanks for taking the time! :)  Some of the quote changes and replacements were RubyMine transformations gone wild and I just hit every single instance of "staging_dir" with it until I could get the unit tests to pass.  I'll fix the text style and comments try to break it up into 2 pull requests.

For fetching a specific SHA - you have to tag it.  You can checkout at a sha just fine - you just need to already be sure that it's reachable from an otherwise tracked remote.  If the warning is too noisy, I can make it an info log or something.

ksubrama:
I pared this pull request down to just the git related fix.  It passes on my mac.  I'll do the Windows related and cosmetic fixes in a separate change.

jaym:
:+1: I restarted the test. Looks like sourceforge might have returned an error

ksubrama:
And @schisamo I just found https://github.com/chef/omnibus/pull/475

So yes we can :)  I'm going to +1 and merge that pull request and then submit the windows test fixes, ensuring that appveyor fails first (in the way I expect it to) and then passes next after my fixes.

PR #522, opened 2015-08-20
Getting unit tests to pass windows.

ksubrama:


spec/functional/file_syncer_spec.rb
```diff
@@ -144,12 +144,17 @@ module Omnibus
...
+            expect("#{destination}/bin/git-write-tree").to be_a_file
+          else
+            expect("#{destination}/bin/git-tag").to be_a_hardlink
+            expect("#{destination}/bin/git-write-tree").to be_a_hardlink
+          end
         end
       end
 
-      context 'with deeply nested paths and symlinks' do
+      context 'with deeply nested paths and symlinks', :not_supported_on_windows do
```
ksubrama: No symlinks on windows

spec/functional/templating_spec.rb
```diff
@@ -51,7 +51,7 @@ class RandomClass
         end
       end
 
-      context 'when a mode is given' do
+      context 'when a mode is given', :not_supported_on_windows do
```
ksubrama: No file modes - not executable anyway.

spec/support/file_helpers.rb
```diff
@@ -39,8 +39,17 @@ def create_file(*paths, &block)
         path
       end
 
-      def create_link(a, b)
-        FileUtils.ln_s(a, b)
+      def create_link(src, dest)
```
ksubrama: Windows and symlinks again

spec/functional/file_syncer_spec.rb
```diff
@@ -144,12 +144,17 @@ module Omnibus
           described_class.sync(source, destination)
 
           expect("#{destination}/bin/git").to be_a_file
-          expect("#{destination}/bin/git-tag").to be_a_hardlink
-          expect("#{destination}/bin/git-write-tree").to be_a_hardlink
+          if windows?
+            expect("#{destination}/bin/git-tag").to be_a_file
```
ksubrama: be_a_hardlink predicate doesn't work on windows.

spec/unit/packagers/base_spec.rb
```diff
@@ -138,19 +142,5 @@ module Omnibus
         end
       end
     end
-
-    describe '#resoures_path' do
```
ksubrama: Duplicate

spec/unit/packagers/makeself_spec.rb
```diff
@@ -44,7 +44,7 @@ module Omnibus
     end
 
     describe '#write_makeselfinst' do
-      it 'generates the executable file' do
+      it 'generates the executable file', :not_supported_on_windows do
```
ksubrama: Executable permissions doesn't make sense on windows

ksubrama:
I've annotated the windows related fixes so that you can ignore the quote changes.  Some of the File.join changes are simply IDE driven and are there to make things uniform.

ksubrama:
Fixes appveyor build introduced in #475

ksubrama:
Source forge is having issues right now.  I'll re-run the tests when it's back up.

scotthain:
This is a more general question - does it make sense to run the packager tests on platforms that don't use that package type? (Like running MSI specs on OSX and vice versa)

Not a big deal, just curious. Otherwise looks ok to me.

ksubrama:
@scotthain Functional tests?  No.  But I'd want to run unit tests on whatever platform I was developing on and not break things.

ksubrama:
I had to disable the net_fetcher functional tests on windows because I couldn't find a sensible tar.exe that accepted windows paths the same way that our omnibus builders do.  I think that's a future task to sort out.  All other tests pass.

jaym:
there should be one in mingw

ksubrama:
That's the one that doesn't seem to work.  See all my previous appveyor build results and appveyor.yml

jaym:
https://github.com/jaym/windows-openssl-build/blob/master/appveyor.yml#L25

ksubrama:
That's the one I copied out of :)

ksubrama:
As an example:  https://ci.appveyor.com/project/chef/omnibus/build/10

           tar (child): Cannot execute remote shell: No such file or directory
       tar (child): C\:\\projects\\omnibus\\spec\\tmp\\software\\local\\cache\\zlib-1.2.8.tar.gz: Cannot open: I/O error
       tar (child): Error is not recoverable: exiting now
       
       gzip: stdin: unexpected end of file
       tar: Child returned status 2
       tar: Exiting with failure status due to previous errors

ksubrama:
@jaym D'oh - actually I screwed up.  I put the devkit tar too late in the path.  It was picking up the wrong tar.  I fixed appveyor.yml to reflect that.

jaym:
:+1: 
In general, I don't think it's worth changing all the " -> ' unless you have something to enforce that invariant

thommay:
:+1:

schisamo:
@ksubrama Let's please remove cc65147 from this PR. We can introduce a style/cosmetic PR once we've decided on a style enforcement tool. I'm personally leaning toward introducing [Finstyle](https://github.com/fnichol/finstyle) which is sanity wrapper around Rubocop. It's an @fnichol joint that is used in [Test Kitchen](https://github.com/test-kitchen/test-kitchen/blob/4c5bb885a9fd5129feb62b470b8ffe1eecbfbafc/test-kitchen.gemspec#L52).

ksubrama:
@schisamo Sounds good.  @jaym thanks for the cert fix.  I forgot that appveyor doesn't do that by default.

PR #523, opened 2015-08-20
Default windows_arch to x86 until Omnibus 5

jaym:
We're going to do an omnibus release, and we'd like to not have to bump the major version

### DIFF
--- a/lib/omnibus/config.rb
+++ b/lib/omnibus/config.rb
@@ -452,11 +452,15 @@ module Omnibus
     #
     # @return [:x86, :x64]
     default(:windows_arch) do
-        if Ohai['kernel']['machine'] == 'x86_64'
-            :x64
-        else
-            :x86
+      if Ohai['kernel']['machine'] == 'x86_64'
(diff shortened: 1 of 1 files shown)


ksubrama:
:+1:

yzl:
:+1:

lib/omnibus/config.rb
```diff
@@ -452,11 +452,15 @@ def reset!
...
     # @return [:x86, :x64]
     default(:windows_arch) do
-        if Ohai['kernel']['machine'] == 'x86_64'
-            :x64
-        else
-            :x86
+      if Ohai['kernel']['machine'] == 'x86_64'
+        Omnibus.logger.deprecated('Config') do
+          "windows_arch is defaulting to :x86. In Omnibus 5, it will " \
+          "default to :x86_64 if the machine architecture is x86_64. " \
```
schisamo: I think you mean `default to :x64`.

schisamo:
Modulo my small comment I am :+1:.

PR #524, opened 2015-08-21
Pin uber-s3 to ~> 0.2.4

jaym:
uber-s3 0.1.1 does not work with the cache commands

### DIFF
--- a/omnibus.gemspec
+++ b/omnibus.gemspec
@@ -27,7 +27,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency 'mixlib-versioning'
   gem.add_dependency 'ohai',             '~> 8.0'
   gem.add_dependency 'ruby-progressbar', '~> 1.7'
-  gem.add_dependency 'uber-s3'
+  gem.add_dependency 'uber-s3',          '~> 0.2.4'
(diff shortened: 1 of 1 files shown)


schisamo:
:+1: 

/cc @chef/omnibus-maintainers

omnibus.gemspec
```diff
@@ -27,7 +27,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency 'mixlib-versioning'
   gem.add_dependency 'ohai',             '~> 8.0'
   gem.add_dependency 'ruby-progressbar', '~> 1.7'
-  gem.add_dependency 'uber-s3'
+  gem.add_dependency 'uber-s3',          '~> 0.2.4'
```
sethvargo: Can we just restrict this to `~> 0.2`? That will be the equivalent of `>= 0.2 && < 1.0`

danielsdeleo: For pre-1.0, I think it's safer to only allow the patch version to float. SemVer allows breaking changes in any pre-1.0 release, but hopefully the authors would at least bump minor.

jaym: @sethvargo are you ok with that?

sethvargo: :+1:

PR #527, opened 2015-08-25
Allow omnibus to build "bundled" Windows installers

thommay:
The delivery-cli package needs to include the MS VC Redist libraries, and so we need to create a bundled installer.
This pull-request implements the logic to do so, and I'd appreciate a review of the approach before I go through and write the necessary tests and so on.

@jaym @ksubrama @schisamo

### DIFF
--- a/lib/omnibus/packagers/msi.rb
+++ b/lib/omnibus/packagers/msi.rb
@@ -30,6 +30,9 @@ module Omnibus
       # Render the source file
       write_source_file
 
+      # Optionally, render the bundle file
+      write_bundle_file if bundle_msi
+
       # Copy all the staging assets from vendored Omnibus into the resources
       # directory.
       create_directory("#{resources_dir}/assets")
@@ -71,7 +74,7 @@ module Omnibus
         # Create the msi, ignoring the 204 return code from light.exe since it is
         # about some expected warnings
 
-        msi_file = windows_safe_path(Config.package_dir, package_name)
+        msi_file = windows_safe_path(Config.package_dir, msi_name)
 
         light_command = <<-EOH.split.join(' ').squeeze(' ').strip
           light.exe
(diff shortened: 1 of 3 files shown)


sethvargo:
This is a bit out of my element, but the code looks good :+1:

jaym:
@thommay why do we want both an msi and a bundle. Is it not pick one?

jaym:
a few unit tests would be nice. Other than that :+1: 
Also, if you have a sample bundle.wxs file, i think it would be good to drop here

thommay:
@jaym you build the msi, and then you chain the MSI into the exe

schisamo:
:+1: Looks useful @thommay!

lib/omnibus/packagers/msi.rb
```diff
@@ -88,6 +91,38 @@ class Packager::MSI < Packager::Base
...
+          bundle_file = windows_safe_path(Config.package_dir, bundle_name)
+
+          bundle_light_command = <<-EOH.split.join(' ').squeeze(' ').strip
+          light.exe
+            -nologo
+            -ext WixUIExtension
+            #{wix_extension_switches(wix_light_extensions)}
+            -cultures:en-us
+            -loc "#{windows_safe_path(staging_dir, 'localization-en-us.wxl')}"
+            project-files.wixobj bundle.wixobj
```
ksubrama: Would a bundle need the set of all project-files again?  I would have thought that the chaining process would only need some kind of reference to the msi.

ksubrama:
http://wixtoolset.org/documentation/manual/v3/bundle/ seems to talk about a bootstrapper application and some new extension to be included.  Can you link to an example bundle.wxs file?  I'm just having a hard time figuring out what gets pulled into what.

thommay:
@ksubrama yeah, previously I was just enabling the extension in the project file, but we can just turn it on across the board. A sample bundle.wxs is https://github.com/chef/delivery-cli/blob/tm/msi_build/omnibus-delivery-cli/resources/delivery-cli/msi/bundle.wxs.erb

thommay:
@jaym @ksubrama @schisamo unit tests and a sample bundle.wxs added. If this looks good I'll squash it up

schisamo:
@thommay This looks great! I'm :+1: after you squash things.

danielsdeleo:
LGTM

thommay:
@schisamo squished!

schisamo:
@thommay Rebase and merge this!

PR #530, opened 2015-08-25
Artifactory publisher improvements

schisamo:
Two small tweaks from Artifactory publishing changes introduced in https://github.com/chef/omnibus/pull/508:

* only `WARN` on missing build platforms during publishing - There are valid cases where publishing will need to be done with a subset of the platform mappings matrix, for example publishing being done directly from a build slave.
* support opting out of build record creation - If we are only publishing for a subset of the platform mappings matrix we will not want an Artifactory build record created as it won’t fully represent all packages in the build.

/cc @chef/omnibus-maintainers @christophermaier

### DIFF
--- a/lib/omnibus/cli/publish.rb
+++ b/lib/omnibus/cli/publish.rb
@@ -62,6 +62,10 @@ module Omnibus
     #
     #   $ omnibus publish artifactory libs-omnibus-local pkg/chef*
     #
+    method_option :build_record,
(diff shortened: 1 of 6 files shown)


yzl:
lgtm

sethvargo:
:+1:

sersut:
Neat! :+1:

seth:
Yay!

PR #533, opened 2015-08-28
Remove uber-s3 dependency for s3_cache

jaym:
This is part 1 of 2 in removing the uber-s3 gem as a dependency. Looks like this gem is no longer being maintained and is causing old gems to be pulled in.

[Part 2](https://github.com/chef/omnibus/pull/534) will remove uber-s3 from the s3 publisher

### DIFF
--- a/lib/omnibus/s3_cache.rb
+++ b/lib/omnibus/s3_cache.rb
@@ -15,7 +15,8 @@
 #
 
 require 'fileutils'
-require 'uber-s3'
+require 'aws-sdk'
+require 'base64'
 
 module Omnibus
   class S3Cache
@@ -41,7 +42,7 @@ module Omnibus
       # @return [Array<String>]
       #
       def keys
-        bucket.objects('/').map(&:key)
+        bucket.objects.map(&:key)
       end
 
       #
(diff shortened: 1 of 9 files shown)


lib/omnibus/s3_cache.rb
```diff
@@ -129,33 +128,13 @@ def key_for(software)
...
-        @bucket ||= begin
-          if client.exists?('/')
-            client.bucket
-          else
-            client.connection.put('/')
-          end
-        end
+      def s3_configuration
+        {
+          region:               'us-east-1',
```
schisamo: Does this need to be configurable? Existing config values live at: 
https://github.com/chef/omnibus/blob/64c477012296f86e9138aaa66ad084f6e2894f21/lib/omnibus/config.rb#L260-L294

schisamo:
:+1: other than my small comment. So nice to be using the 'official' aws library!

lib/omnibus/s3_cache.rb
jaym: we can make it configurable. IIRC, us-east-1 will forward to the correct region, so we can use that as a fallback.

yzl:
lgtm, pending making regions configurable

yzl:
:+1:

schisamo:
:+1:

PR #534, opened 2015-08-28
Modify s3 publisher to use aws-sdk instead of uber-s3

jaym:


lib/omnibus/publishers/s3_publisher.rb
```diff
@@ -45,18 +45,13 @@ def publish(&block)
...
-    # The actual S3 client object to communicate with the S3 API.
-    #
-    # @return [UberS3]
-    #
-    def client
-      @client ||= UberS3.new(
-        access_key:        Config.publish_s3_access_key,
+    def s3_configuration
+      {
+        region:            'us-east-1',
```
schisamo: Similar to my comment from #533. Do we need to make this configurable as part of:
https://github.com/chef/omnibus/blob/64c477012296f86e9138aaa66ad084f6e2894f21/lib/omnibus/config.rb#L377-L397

yzl:
lgtm, pending comments

yzl:
:+1:

schisamo:
:+1:

PR #536, opened 2015-08-28
Add architecture to package name on windows

jaym:
Potentially breaking change?

cc @chef/omnibus-maintainers @ksubrama @chefsalim @thommay

### DIFF
--- a/lib/omnibus/packagers/msi.rb
+++ b/lib/omnibus/packagers/msi.rb
@@ -321,11 +321,11 @@ module Omnibus
     end
 
     def msi_name
-      "#{project.package_name}-#{project.build_version}-#{project.build_iteration}.msi"
+      "#{project.package_name}-#{project.build_version}-#{project.build_iteration}-#{Config.windows_arch}.msi"
     end
 
     def bundle_name
-      "#{project.package_name}-#{project.build_version}-#{project.build_iteration}.exe"
+      "#{project.package_name}-#{project.build_version}-#{project.build_iteration}-#{Config.windows_arch}.exe"
     end
 
     #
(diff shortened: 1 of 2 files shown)


jaym:
dont' know if we want this in 4.1

lib/omnibus/packagers/msi.rb
```diff
@@ -321,11 +321,11 @@ def package_name
     end
 
     def msi_name
-      "#{project.package_name}-#{project.build_version}-#{project.build_iteration}.msi"
+      "#{project.package_name}-#{project.build_version}-#{project.build_iteration}.#{Config.windows_arch}.msi"
```
schisamo: Looking at how other Windows packages are named ([Ruby](http://rubyinstaller.org/downloads/), [Git](https://git-scm.com/download/win) and [various Chocolatey stuff](https://chocolatey.org/packages?q=64-bit)) I think we should separate iteration and arch with a dash (`-`).

schisamo:
I'm not sure I would call this a breaking change but my arm could be twisted if a good case is made.

lib/omnibus/packagers/msi.rb
sethvargo: :+1:

sethvargo:
I think this is fine to put into 4.1 as long as it's called out in the changelog. It isn't breaking any APIs, just the output format. As long as it's clearly called out in the CL, I think it's fine.

ksubrama:
:+1:

yzl:
:+1:

jaym:
cc @chef/omnibus-maintainers

schisamo:
:+1:

PR #535, opened 2015-08-28
Use windows_arch to get arch in metadata

jaym:
Windows is special. A builder should not rely
on it's native architecture.

cc @ksubrama @chefsalim @schisamo

PR #537, opened 2015-08-29
`Metadata#arch` should return a valid Ohai architecture

schisamo:
The change introduced in #535 began returning architecture values of `:x86` and `:x64` on Windows. We would prefer the `*.metadata.json` files contain valid Ohai-based values of `i386` or `x86_64` as we use this data when publishing to backend systems like Artifactory. This saves unnecessary translation when fetching packages at install time.

/cc @chef/engineering-services @chef/omnibus-maintainers @chef/client-windows

### DIFF
--- a/lib/omnibus/metadata.rb
+++ b/lib/omnibus/metadata.rb
@@ -108,8 +108,8 @@ module Omnibus
       # @return [String]
       #
       def arch
-        if Ohai['platform'] == 'windows'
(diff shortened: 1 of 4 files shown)


jaym:
:+1:

sethvargo:
LGTM

stevendanna:
:+1:

PR #538, opened 2015-08-31
Release Omnibus 4.1.0

jaym:
Any objections to me doing this today?

cc @chef/omnibus-maintainers


schisamo:
We should ensure https://github.com/chef/omnibus/pull/536 makes the cut.

schisamo:
:joy_cat:

PR #539, opened 2015-09-01
initial Wind River Linux 5/7 support for Cisco Nexus devices

mattray:
Add wrlinux platform and nexus platform_family support for building on Wind
River Linux 5.


lib/omnibus/packagers/rpm.rb
```diff
@@ -494,7 +495,11 @@ def safe_version
       #   http://rpm.org/ticket/56
       #
       if version =~ /\-/
-        converted = version.gsub('-', '~')
+        if Ohai['platform_family'] == 'wrlinux'
```
schisamo: We should find a way to make this check more general. Should't it also be based on a combination of platform and platform version?

resources/rpm/spec.erb
```diff
@@ -19,7 +19,9 @@ Name: <%= name %>
 Version: <%= version %>
 Release: <%= iteration %><%= dist_tag %>
 Summary:  <%= description.split("\n").first.empty? ? "_" : description.split("\n").first %>
+-<% if platform != 'nexus' -%>
 BuildArch: <%= architecture %>
+-<% end -%>
```
schisamo: Why is this excluded?

lib/omnibus/health_check.rb
```diff
@@ -51,6 +51,17 @@ class HealthCheck
       /libutil\.so/,
     ].freeze
 
+    WRLINUX_WHITELIST_LIBS = [
+      /libc\.so/,
+      /libcrypt\.so/,
+      /libdl\.so/,
+      /libffi\.so/,
```
schisamo: We compile `libffi`. Excluding it from the health check is dangerous. Everything else already exists in the default `WHITELIST_LIBS `.

schisamo:
This also needs some test coverage.

lib/omnibus/packagers/rpm.rb
mattray: Right now we only have WRL5 but I can update this when I add support for WRL7 if necessary.

resources/rpm/spec.erb
mattray: The old version of RPM didn't like that being in the RPM spec.

mattray:
Updated omnibus-software to fix libffi on Wind River Linux.
https://github.com/chef/omnibus-software/pull/480

lib/omnibus/health_check.rb
```diff
@@ -51,6 +51,16 @@ class HealthCheck
...
+    WRLINUX_WHITELIST_LIBS = [
+      /libc\.so/,
+      /libcrypt\.so/,
+      /libdl\.so/,
+      /libm\.so/,
+      /libpthread\.so/,
+      /librt\.so/,
+      /libutil\.so/,
+    ].freeze
+
```
schisamo: All of these libraries exist in the default `WHITELIST_LIBS`...we shouldn't need to maintain a special list for Wind River.

resources/rpm/spec.erb
schisamo: This feels related to https://github.com/chef/omnibus/pull/542. We might not need `BuildArch` in the spec file.

schisamo:
Left a few more comments. Reminder we also need some test coverage before this will be merged.

mattray:
@schisamo @chef/omnibus-maintainers merged in @adamleff's support for Wind River Linux 7, which also brings in pending tests as we wait for upstream fauxhai patches to be merged.

spec/unit/metadata_spec.rb
```diff
@@ -80,7 +80,9 @@ module Omnibus
       it_behaves_like 'a version manipulator', 'fedora', '11.5', '11'
       it_behaves_like 'a version manipulator', 'freebsd', '10.0', '10'
       it_behaves_like 'a version manipulator', 'gentoo', '2004.3', '2004.3'
+      it_behaves_like 'a version manipulator', 'ios_xr', '6.0.0.14I', '6'
       it_behaves_like 'a version manipulator', 'mac_os_x', '10.9.1', '10.9'
+      it_behaves_like 'a version manipulator', 'nexus', '5.0', '5'
```
glennmatthews: This looks suspect to me. Line 83 is dealing with the IOS XR release version (6.0.0.14I) rather than the WRL 7 release version (7.0.0.2), but line 85 is dealing with the WRL 5 release version (5.0) rather than the NX-OS release version (7.0(3)I2(1)). Which way should it be?

adamleff: @glennmatthews it doesn't really matter as this is just a unit test ensuring that whatever version is passed in is stripped down to its major version number for both "nexus" and "ios_xr" platforms.  I could have passed in 10.0.1.192123 as a version number for this unit under test, and provided it still returned 10, the test is operating properly.

With regards to the versions themselves, Ohai will be overwriting the VERSION field from /etc/os-release with /etc/cisco-release on WRL7, so the version from /etc/cisco-release will take precedence, which is why I used 6.0.0.14I in my test above.  On the WRL5 builder image I was provided, the VERSION field in /etc/os-release was simply "5" so I used "5.0" in my test above to ensure the string was truncated down to 5 as we expect.

adamleff:
@schisamo @chef/omnibus-maintainers I believe this is ready for merge. Can we get some :+1:'s please?

yzl:
:+1:

schisamo:
@adamleff Looks like this branch needs a rebase before merging.

adamleff:
@schisamo it didn't earlier today :smile:

@mattray: since it's your branch, can you rebase, please?

mattray:
@schisamo @yzl @adamleff rebased, thanks in advance for merging!

PR #541, opened 2015-09-03
Fix missing package name in signature

chefsalim:
When installing signed packages on Windows, the UAC warning currently gives a random "Program name" in the warning.  This change adds a fix so that we set the package name when signing, which then allows Windows to display the correct name.

cc @schisamo @jaym @sersut


schisamo:
@chefsalim be sure to /cc @chef/omnibus-maintainers .

schisamo:
:+1:

jaym:
:+1:

adamedx:
Has this always been broken, or something recently regressed?

chefsalim:
Looks like it has always been broken - DK 0.7 had the same behavior.

yzl:
:+1:

PR #550, opened 2015-09-21
support epoch in RPM packages

chaiken-verticloud:
Epoch number is required for some RPM packages to work correctly.

I considered adding equivalent support for Debian packages, but Omnibus already supports Debian packages by allowing a colon in the version number.

### DIFF
--- a/spec/unit/packagers/rpm_spec.rb
+++ b/spec/unit/packagers/rpm_spec.rb
@@ -91,6 +91,20 @@ module Omnibus
       end
     end
 
+    describe '#epoch' do
+      it 'is a DSL method' do
+        expect(subject).to have_exposed_method(:epoch)
+      end
+
+      it 'has no default value' do
+        expect(subject.epoch).to eq(nil)
+      end
+
+      it 'must be a Fixnum' do
+        expect { subject.epoch(Object.new) }.to raise_error(InvalidValue)
+      end
+    end
+
     describe '#category' do
       it 'is a DSL method' do
         expect(subject).to have_exposed_method(:category)
@@ -156,6 +170,14 @@ module Omnibus
         expect(contents).to include("URL: https://example.com")
         expect(contents).to include("Packager: Chef Software")
         expect(contents).to include("Obsoletes: old-project")
+      end
+
+      it 'supports the epoch tag' do
+        allow(subject).to receive(:epoch).and_return(42)
+        subject.write_rpm_spec
+        contents = File.read(spec_file)
+
+        expect(contents).to include("Epoch: 42")
       end
 
       context 'when scripts are given' do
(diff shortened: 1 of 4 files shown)


spec/unit/packagers/rpm_spec.rb
```diff
@@ -158,6 +172,14 @@ module Omnibus
         expect(contents).to include("Obsoletes: old-project")
       end
 
+      it 'supports the epoch tag' do
```
stevendanna: Can we also add a test for the case when an epoch isn't exlicitly set to ensure that the spec doesn't include the Epoch: line?

stevendanna:
@chaiken-verticloud Apologies for the delay in this.  This looks good to me but I'd like to see an additional test to cover the case of no epoch being set.

cc @chef/omnibus-maintainers

yzl:
Code looks ok to me, although I agree with @ssd about the need for a test.  After reading the [recommendations in the rpm manual](http://www.rpm.org/max-rpm-snapshot/s1-rpm-depend-manual-dependencies.html), I am :-1: about explicit support for epoch, but will defer to the consensus of other maintainers.

stevendanna:
@yzl Mind expanding a bit on why? I don't disagree, just think it would help us gain consensus.

The document you linked to says:

>  If you have the option between changing the software's version-numbering scheme, or using epoch numbers in RPM, please consider changing the version-numbering scheme. Chances are, if RPM can't figure it out, most of the people using your software can't, either. But in case you aren't the author of the software you're packaging, and its version numbering scheme is giving RPM fits, the epoch tag can help you out. 

Since this relates to the version you give to omnibus packages, I think in most cases we'd prefer users go with the "change your version numbering scheme" solution. I guess the edge case are users who are trying to faithfully package some upstream package using omnibus and want to keep version numbers in line.

yzl:
@ssd My reasons, which I admit are weird and idiosyncratic, are:

1.  The non-enthusiasm in the rpm docs I linked to in my previous comment.
2.  I see epoch as enough of an edge case that I'd rather that projects that need to use it [override the spec file locally](https://github.com/chef/omnibus/blob/master/lib/omnibus/packagers/base.rb#L184-L198).  This is less nice than having `omnibus` support, but that's the nature of edge cases.

scotthain:
I'm :+1: on what @yzl said, for similar weird and idiosyncratic reasons...

chaiken-verticloud:
Unfortunately, I converted an fpm-packaged gem into an omnibus-packaged gem.  Due to oddities in the way that fpm worked, I need to change the epoch.  This change seems to be a pretty innocuous, backward-compatible addition of a small feature.  It was pretty easy for me to merge into master as of today, which leads me to believe that it is also well-contained and easy to maintain over time.

PR #551, opened 2015-09-23
Add info message for publish cli corner case

scotthain:
Simple corner case on the off chance you are using publish from command line, if no packages are found, it silently 'fails'

@chef/omnibus-maintainers

### DIFF
--- a/lib/omnibus/publishers/artifactory_publisher.rb
+++ b/lib/omnibus/publishers/artifactory_publisher.rb
@@ -23,6 +23,10 @@ module Omnibus
       log.info(log_key) { 'Starting artifactory publisher' }
       safe_require('artifactory')
 
+      if packages.empty?
+        log.info(log_key) { 'No packages found, skipping publish' }
+      end
+
       packages.each do |package|
         # Make sure the package is good to go!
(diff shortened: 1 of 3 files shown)


lib/omnibus/publishers/artifactory_publisher.rb
```diff
@@ -23,6 +23,10 @@ def publish(&block)
       log.info(log_key) { 'Starting artifactory publisher' }
       safe_require('artifactory')
 
+      if packages.empty?
+        log.info(log_key) { 'No packages found, skipping publish' }
+      end
+
```
schisamo: We should probably put this message up in `Omnibus::Publisher#packages`:
https://github.com/chef/omnibus/blob/shain/add_publish_debug/lib/omnibus/publisher.rb#L68-L117

This will ensure all publishers benefit.

schisamo:
One comment, we should also add some test coverage.

scotthain:
Test added!

schisamo:
:+1:

yzl:
:+1:

curiositycasualty:
🚀🍰

PR #554, opened 2015-10-06
Use the proper architecture value in Solaris packages

schisamo:
We should be using the more general instruction set architecture (e.g. `sparc` vs `sun4v`/`sun4u`). I also added missing test coverage for the Solaris packager!

Fixes chef/chef#4018

/cc @chef/omnibus-maintainers @chef/engineering-services @lamont-granquist @MarkGibbons

### DIFF
--- a/lib/omnibus/packagers/solaris.rb
+++ b/lib/omnibus/packagers/solaris.rb
@@ -14,24 +14,27 @@
 # limitations under the License.
 #
 
+require 'socket'
+
 module Omnibus
   class Packager::Solaris < Packager::Base
+    # @return [Hash]
+    SCRIPT_MAP = {
+      # Default Omnibus naming
+      postinst:  'postinstall',
+      postrm: 'postremove',
+      # Default Solaris naming
+      postinstall:  'postinstall',
+      postremove: 'postremove',
+    }.freeze
+
     id :solaris
 
     build do
-      shellout! "cd #{install_dirname} && find #{install_basename} -print > #{staging_dir_path('files')}"
-      
-
-      write_prototype_content
-
-      write_pkginfo_content
-
-      copy_file("#{project.package_scripts_path}/postinst", staging_dir_path('postinstall'))
-      copy_file("#{project.package_scripts_path}/postrm", staging_dir_path('postremove'))
-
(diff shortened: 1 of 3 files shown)


scotthain:
I hate that travis timing issue on the datestamp test - may want to update the test to regex ignore the seconds? Other than that - code looks good! :+1:

spec/unit/packagers/solaris_spec.rb
```diff
@@ -0,0 +1,231 @@
...
+        expect(contents).to include("BASEDIR=/opt")
+        expect(contents).to include("PKG=project")
+        expect(contents).to include("NAME=project")
+        expect(contents).to include("ARCH=i386")
+        expect(contents).to include("VERSION=1.2.3-1")
+        expect(contents).to include("CATEGORY=application")
+        expect(contents).to include("DESC=")
+        expect(contents).to include("VENDOR=Chef Software")
+        expect(contents).to include("EMAIL=Chef Software")
+        expect(contents).to include("PSTAMP=#{`hostname`.chomp + Time.now.utc.iso8601}")
```
danielsdeleo: You probably want to stub Time here, travis executed the test on a second boundary and failed.

danielsdeleo:
LGTM modulo the time stubbing thing. :+1:

lib/omnibus/packagers/solaris.rb
```diff
@@ -88,17 +111,44 @@ def write_pkginfo_content
...
+    end
+
+    #
+    # The architecture for this Solaris package.
+    #
+    # @return [String]
+    #
+    def safe_architecture
+      # The #intel? and #sparc? helpers come from chef-sugar
+      if intel?
```
jaym: why not `i386?`?

lamont-granquist:
LGTM :+1:

lib/omnibus/packagers/solaris.rb
schisamo: Because there is a bug in chef-sugar:
https://github.com/sethvargo/chef-sugar/pull/121

Without that fix this spec fails:
https://github.com/chef/omnibus/blob/074002633760d29d86d5c651768493e7c00e666d/spec/unit/packagers/solaris_spec.rb#L225-L231

PR #555, opened 2015-10-06
Ensure metadata.json returns the correct Solaris arch value

schisamo:
This is an additional fix for chef/omnibus#554.

/cc @chef/omnibus-maintainers @chef/engineering-services  @lamont-granquist @jkeiser @jaym


scotthain:
I am vaguely concerned about the i386 etc thing. I don't want to overload what we consider in opscode-ci i386 as 32bit. The problem with Solaris is it's 64bit but ohai sets it as i386... Let's talk more about this.

schisamo:
@scotthain IMO we need to decide if `Omnibus::Metdata` returns:

1. The Ohai value for arch: `i86pc`, `sun4v`, `sun4u` etc.
2. The same arch as the  `Omnibus::Packager::Solaris` value as dictated by `pkginfo` and the [Solaris packaging standards](http://docs.oracle.com/cd/E19683-01/816-0219/6m6njqbat/index.html): 

> ARCH values from the following Solaris-supported instruction set architectures (uname -p): sparc, i386, and ppc

For sanity's sake I think we should go with `#2` which ensures a `*.solaris` package and its `*.solaris.metdata.json` file are in sync WRT to architecture.

stevendanna:
:+1: I am for (2).  I think that makes the most sense in the context of an omnibus package.

lamont-granquist:
:+1: for 2

scotthain:
I'm fine with 2 :+1:

PR #556, opened 2015-10-06
Change the appbundler DSL to not make an apps dir

ksubrama:
Also enable functional tests on windows.

### DIFF
--- a/lib/omnibus/builder.rb
+++ b/lib/omnibus/builder.rb
@@ -276,27 +276,12 @@ module Omnibus
     def appbundle(app_name, options = {})
       build_commands << BuildCommand.new("appbundle `#{app_name}'") do
         bin_dir            = "#{install_dir}/bin"
-        embedded_apps_root = "#{install_dir}/embedded/apps"
-        embedded_app_dir   = "#{embedded_apps_root}/#{app_name}"
-        gemfile_lock       = "#{embedded_app_dir}/Gemfile.lock"
         appbundler_bin     = windows_safe_path("#{install_dir}/embedded/bin/appbundler")
 
         # Ensure the main bin dir exists
         FileUtils.mkdir_p(bin_dir)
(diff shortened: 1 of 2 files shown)


sethvargo:
Code looks good, but what is the motivation?

jaym:
I feel like this will break windows upgrades for chef

danielsdeleo:
Don't the tests rely on having the full source code in the apps/ directory?

ksubrama:
@jaym It absolutely will - but not for the service. It's break for the ETW dll we register.  I am making changes to the chef repo itself to include that dll in the chef gem on windows so that our event registration code can point to that instead.  The service is ok because it points to a path in opscode/chef/bin which will correctly point to the gem (that was your fix for 12.4.3 right?).

ksubrama:
@sethvargo This was our attempt to not include multiple copies of our apps because there was an incident where a chef plugin somehow ended up including a file from the chef gem instead of the copy of chef from the apps directory which caused a conflicting redefinition of a module.

ksubrama:
@danielsdeleo The full source tree is already in the gem directory.  We'd just have to change the test scripts to look in there instead.

danielsdeleo:
That's fine, but we should have that change queued up so we're not busted in the interim.

ksubrama:
Changed to omnibus software https://github.com/chef/omnibus-software/pull/500

ksubrama:
Changed to omnibus-chef https://github.com/chef/omnibus-software/pull/500

danielsdeleo:
FYI, the "verify" code in chef-dk seems to assume the presence of the apps/ dir quite heavily: https://github.com/chef/chef-dk/blob/master/lib/chef-dk/command/verify.rb

ksubrama:
Changes to chef-dk https://github.com/chef/chef-dk/pull/562

PR #558, opened 2015-10-19
use the wixutil extension

mwrock:
This extensin includes the following custom actions:
```
WixBroadcastSettingChange
WixBroadcastEnvironmentChange
```
Including these actions (separate PR to omnibus-chef) properly broadcast environment variable changes and fixes https://github.com/chef/chef-dk/issues/534

### DIFF
--- a/lib/omnibus/packagers/msi.rb
+++ b/lib/omnibus/packagers/msi.rb
@@ -80,6 +80,7 @@ module Omnibus
           light.exe
             -nologo
             -ext WixUIExtension
+            -ext WixUtilExtension
             #{wix_extension_switches(wix_light_extensions)}
             -cultures:en-us
             -loc "#{windows_safe_path(staging_dir, 'localization-en-us.wxl')}"


schisamo:
@mwrock Be sure to always /cc @chef/omnibus-maintainers

schisamo:
I would only make this change if we want *all* MSIs produced by Omnibus to do this. If it's just something we want in a single project (like ChefDK or Chef) we should be able to fix this in a project definition using the `wix_light_extension` DSL method for the MSI packager with something like this:

```ruby
package :msi do
  upgrade_code "AB1D6FBD-F9DC-4395-BDAD-26C4541168E7"
  signing_identity "F74E1A68005E8A9C465C3D2FF7B41F3988F0EA09", machine_store: true
  wix_light_extension "WixUtilExtension"
end
```

mwrock:
That's reasonable. This was made solely for dk and I'll figure out where that above snippet needs to live.

mwrock:
I see where thats done. adding to omnibus-chef PR and running another test to validate.

PR #559, opened 2015-10-20
Skip registry harvesting, suppress warning

chefsalim:
Heat currently spits out tons of warnings like this:
| heat.exe : warning HEAT5150 : Could not harvest data from a file that was expected to be a SelfReg DLL: C:\opscode\chef\embedded\bin\libeay32.dll. If this file does not support SelfReg you can ignore this warning. Otherwise, this error detail may be helpful to diagnose the failure: Unable to load file: C:\opscode\chef\embedded\bin\libeay32.dll, error: 193

These are false positives.  Since we don't actually do any registry harvesting, we can turn these off by passing the "-sreg" flag to heat.

@chef/omnibus-maintainers @ksubrama


schisamo:
:+1:

PR #563, opened 2015-11-03
[WIP] FastMSI Packager

chefsalim:
:construction: 

This PR adds support for a new 'FastMSI' packager for Windows. This packager uses a strategy of creating a zip file, that will then get unzipped during install, rather than having Windows Installer track every file separately. This significantly decreases the time required to install the package on Windows.

This is the first of two PRs. The second PR will be for omnibus-chef which will activate and start making use of this infra.  By itself, this PR will not cause any changes to the output of current builds, and so should be mostly a no-op.

One functional change is with how the packager is being created. There is a new packager_override flag for projects, that can be used to override the system default packager.  This is only used for the Windows platform currently.

@chef/omnibus-maintainers @btm @adamedx @mwrock @ksubrama @jaym


lib/omnibus/packagers/fastmsi.rb
```diff
@@ -0,0 +1,654 @@
...
+      # Zip up the contents of the install directory
+      zip_command = <<-EOH.split.join(' ').squeeze(' ').strip
+        7z a -r
+          #{windows_safe_path(staging_dir)}\\#{project.name}.zip
+          #{windows_safe_path(project.install_dir)}\\*
+      EOH
+      shellout!(zip_command, returns: [0])
+
+      # Harvest the zip file with heat.exe
+      Dir.chdir(staging_dir) do
```
ksubrama: I don't think you need heat to produce the xml entry for one zip file.  It might actually be much clearer to simply point to the zip file in sources.wxs and do away with project-files.wxs.

lib/omnibus/packagers/fastmsi.rb
```diff
@@ -0,0 +1,654 @@
...
+
+        # Compile with candle.exe
+        log.debug(log_key) { "wix_candle_flags: #{wix_candle_flags}" }
+
+        candle_command = <<-EOH.split.join(' ').squeeze(' ').strip
+          candle.exe
+            -nologo
+            #{wix_candle_flags}
+            #{wix_extension_switches(wix_candle_extensions)}
+            -dProjectSourceDir="#{windows_safe_path(project.install_dir)}" "project-files.wxs"
```
ksubrama: No ProjectSourceDir anymore.  If we have a zip, we shouldn't be referring to source files.

lib/omnibus/packagers/fastmsi.rb
```diff
@@ -0,0 +1,654 @@
...
+    #
+    # Write the source file into the staging directory.
+    #
+    # @return [void]
+    #
+    def write_source_file
+      paths = []
+
+      # Remove C:/
+      install_dir = project.install_dir.split('/')[1..-1].join('/')
```
ksubrama: I don't think this is relevant any more either.  No directory hierarchy to pre-create.

resources/fastmsi/source.wxs.erb
```diff
@@ -0,0 +1,73 @@
...
+    <Upgrade Id="$(var.UpgradeCode)">
+      <UpgradeVersion OnlyDetect="yes" Minimum="$(var.VersionNumber)" IncludeMinimum="no" Property="NEWERVERSIONDETECTED" />
+      <UpgradeVersion Minimum="0.0.0.0" IncludeMinimum="yes" Maximum="$(var.VersionNumber)" IncludeMaximum="yes" Property="OLDERVERSIONBEINGUPGRADED" />
+    </Upgrade>
+
+    <InstallExecuteSequence>
+      <RemoveExistingProducts After="InstallValidate" />
+    </InstallExecuteSequence>
+
+    <Directory Id="TARGETDIR" Name="SourceDir">
```
ksubrama: Ditto - this isn't needed any more.

lib/omnibus/packagers/fastmsi.rb
```diff
@@ -0,0 +1,654 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'pathname'
+
```
ksubrama: Can you leave a not at the top of packages/msi and this file, saying that you copy-pastad the code for the purposes of an experiment and that fixes need to go into both of them.  Hopefully, this will succeed and we can kill the original msi packager completely.

lib/omnibus/packagers/fastmsi.rb
chefsalim: Yeah I considered removing it but left it in for now just in the off chance we may need to add some other files (unknown unknowns) as we refine the install.  Will remove as part of refactoring (separate card) if we don't need it by then.

lib/omnibus/packagers/fastmsi.rb
chefsalim: Yep same as above.

lib/omnibus/packagers/fastmsi.rb
chefsalim: Will add.

curiositycasualty:
This looks really promising :+1:

schisamo:
@chefsalim I see *A LOT* of code duplication with `Omnibus::Packager::MSI`. Can we make `Omnibus::Packager::FastMSI` just extend `Omnibus::Packager::MSI` and override a few select methods? Another option would to just expose a single config option on `Omnibus::Packager::MSI` to turn on this monolithic zip, fast approach.

chefsalim:
@schisamo Yes, agreed - we are planning a separate card for refactoring and eliminating the duplication.

mwrock:
Maybe we should mark this pr as WIP and work that card on this PR

schisamo:
Great idea @mwrock. We usually use :construction: :construction: :construction: (along with WIP in the title).

PR #565, opened 2015-11-05
Add fastmsi infra to omnibus

chefsalim:
This PR adds support for a new 'Fast MSI' mode to the Windows MSI packager. This allows the MSI packager to use a strategy of creating a zip file, that will then get unzipped during install, rather than having Windows Installer track every file separately. This significantly (50-60% for a large package like ChefDK) decreases the time required to install the package on Windows.

This is the first of two PRs. The second PR will be for omnibus-chef which will activate and start making use of this infra. By itself, this PR will not cause any changes to the output of current builds, and so should be mostly a no-op.

@chef/omnibus-maintainers @btm @adamedx @mwrock @ksubrama @jaym @schisamo


spec/unit/packagers/msi_spec.rb
```diff
@@ -316,6 +316,23 @@ module Omnibus
       end
     end
 
+    describe '#fast_msi' do
```
mwrock: Seems like there may be alot of tests in this spec that could be influenced by fast_msi. Would it make sense to create shared examples and run through both implementations?

mwrock:
I'm generally :+1: but I'd think we could make the separation of `msi` and `fast_msi` cleaner by having them in separate classes where `fast_msi` either derrives from `msi` or both derrive from a `base_msi`

chefsalim:
Given the way that the packagers are being instantiated, creating a different sub-class isn't quite that straightforward and we have to deal with more new issues and more duplicated code. Perhaps we could have different helper classes under MSI. Eventually we could also just get rid of the regular MSI behavior completely.

spec/unit/packagers/msi_spec.rb
chefsalim: The primary path that fastmsi impacts is functional - the setup and build steps, which don't currently have tests. If we had those yes we could run both modes.

mwrock: ok. yeah. looking at it more closely i see they don't call the `build` method which is where the bulk of the fastmsi stuff is.

mwrock:
ok. my :+1: still stands.

lib/omnibus/packagers/msi.rb
```diff
@@ -47,24 +47,46 @@ class Packager::MSI < Packager::Base
       FileSyncer.glob("#{resources_path}/assets/*").each do |file|
         copy_file(file, "#{resources_dir}/assets/#{File.basename(file)}")
       end
+
+      # If fastmsi, copy externally built custom action DLL into the staging directory
```
stevendanna: Let's drop this comment, code is straightforward enough to not require it.

lib/omnibus/packagers/msi.rb
```diff
@@ -47,24 +47,46 @@ class Packager::MSI < Packager::Base
...
+        else
+          heat_command = <<-EOH.split.join(' ').squeeze(' ').strip
+            heat.exe dir "#{windows_safe_path(project.install_dir)}"
+              -nologo -srd -sreg -gg -cg ProjectDir
+              -dr PROJECTLOCATION
+              -var "var.ProjectSourceDir"
+              -out "project-files.wxs"
+          EOH
+        end
+        shellout!(heat_command, returns: [0])
```
stevendanna: It might be nice to make `heat_command` a method that returns the relevant command as this build block is getting rather long.

lib/omnibus/packagers/msi.rb
```diff
@@ -47,24 +47,46 @@ class Packager::MSI < Packager::Base
...
         copy_file(file, "#{resources_dir}/assets/#{File.basename(file)}")
       end
+
+      # If fastmsi, copy externally built custom action DLL into the staging directory
+      copy_file("#{resources_path}/ext/CustomActionFastMsi.CA.dll", "#{staging_dir}") if fast_msi
     end
 
     build do
+      # If fastmsi, zip up the contents of the install directory
+      zip_command = <<-EOH.split.join(' ').squeeze(' ').strip
```
stevendanna: Similar to my comment on heat_command, I would consider making zip_command a method that returns this string.

stevendanna:
Left some minor ruby style suggestions, leaving the substantive :+1:s to our Windows folks

lib/omnibus/packagers/msi.rb
```diff
@@ -47,24 +47,46 @@ class Packager::MSI < Packager::Base
       FileSyncer.glob("#{resources_path}/assets/*").each do |file|
         copy_file(file, "#{resources_dir}/assets/#{File.basename(file)}")
       end
+
+      # If fastmsi, copy externally built custom action DLL into the staging directory
+      copy_file("#{resources_path}/ext/CustomActionFastMsi.CA.dll", "#{staging_dir}") if fast_msi
```
schisamo: This DLL should ship as part of the Omnibus framework (in the gem's `resources/msi` directory). It can be copied into the staging directory like so:

```ruby
copy_file(resource_path('CustomActionFastMsi.CA.dll'), staging_dir)
```

schisamo:
@chefsalim See my comment above about packaging `CustomActionFastMsi.CA.dll` into this gem. We also need to ensure the default [source.wxs.erb template](https://github.com/chef/omnibus/blob/master/resources/msi/source.wxs.erb) is updated with the changes you added in chef/omnibus-chef#540. This template is used at [new project generation time](https://github.com/chef/omnibus/blob/master/lib/omnibus/generator.rb#L127-L140) AND [MSI creation time](https://github.com/chef/omnibus/blob/master/lib/omnibus/packagers/base.rb#L184-L209) (if a `resources/<PROJECT>/msi/source.wxs.erb` file does not exist in the Omnbius project directory).

chefsalim:
@schisamo @stevendanna @mwrock changes made re: your feedback.

lib/omnibus/packagers/msi.rb
```diff
@@ -47,48 +47,26 @@ class Packager::MSI < Packager::Base
       FileSyncer.glob("#{resources_path}/assets/*").each do |file|
         copy_file(file, "#{resources_dir}/assets/#{File.basename(file)}")
       end
+
+      copy_file(resource_path('CustomActionFastMsi.CA.dll'), staging_dir) if fast_msi
```
schisamo: We should add some comments mentioning how this `*.dll` is built and where the source resides (I know you mentioned https://github.com/chef/fastmsi-custom-action on another PR).

spec/unit/packagers/msi_spec.rb
```diff
@@ -316,6 +336,23 @@ module Omnibus
...
+          subject.fast_msi(Object.new)
+        }.to raise_error(InvalidValue)
+      end
+
+      it 'returns the given value' do
+        subject.fast_msi(true)
+        expect(subject.fast_msi).to be_truthy
+      end
+    end
+
```
schisamo: We need to add some unit test coverage for `Omnibus::Packager::MSI#zip_command` and `Omnibus::Packager::MSI#heat_command`. We need to verify the correct version of the command is returned when `Omnibus::Packager::MSI#fast_msi` is true AND that the command parameters are correct.

schisamo:
@chefsalim Added a few more comments but overall this is looking much closer to what I was envisioning...thank you for taking the time to refactor things! :ice_cream:

lib/omnibus/packagers/msi.rb
```diff
@@ -96,32 +74,11 @@ class Packager::MSI < Packager::Base
...
-          shellout! <<-EOH.split.join(' ').squeeze(' ').strip
-          candle.exe
-            -nologo
-            #{wix_candle_flags}
-            -ext WixBalExtension
-            #{wix_extension_switches(wix_candle_extensions)}
-            -dOmnibusCacheDir="#{windows_safe_path(File.expand_path(Config.cache_dir))}"
-            "#{windows_safe_path(staging_dir, 'bundle.wxs')}"
-          EOH
+          shellout!(candle_command(true))
```
btm: It'd be nice to be able to use keyword arguments here so we know what true means without having to go look for the method. I've submitted #567 to bump the minimum ruby version for omnibus to 2.0 since we're using 2.1.5 anyway.

```
shellout!(candle_command(is_bundle: true))

def candle_command(is_bundle: false)
  # blah
end
```

resources/msi/source.wxs.erb
```diff
@@ -12,23 +12,57 @@
           Version="$(var.VersionNumber)" Manufacturer="!(loc.ManufacturerName)" UpgradeCode="$(var.UpgradeCode)">
 
     <!--
-      Define the minimum supported installer version (2.0).
-      The install should be done for the whole machine not just the current user
+      Minimum installer version (5.0) - Window 7/Server 2008 R2 and above.
+      The install scope is per machine, not the current user
     -->
-    <Package InstallerVersion="200" InstallPrivileges="elevated" Compressed="yes" InstallScope="perMachine" />
+    <Package InstallerVersion="500" InstallPrivileges="elevated"
```
btm: Are we bumping this for a feature we need from the installer or based on our supported platforms?

resources/msi/source.wxs.erb
```diff
@@ -12,23 +12,57 @@
...
     <Media Id="1" Cabinet="Project.cab" EmbedCab="yes" CompressionLevel="high" />
 
-    <!-- Major upgrade -->
-    <Upgrade Id="$(var.UpgradeCode)">
-      <UpgradeVersion OnlyDetect="yes" Minimum="$(var.VersionNumber)" IncludeMinimum="no" Property="NEWERVERSIONDETECTED" />
-      <UpgradeVersion Minimum="0.0.0.0" IncludeMinimum="yes" Maximum="$(var.VersionNumber)" IncludeMaximum="yes" Property="OLDERVERSIONBEINGUPGRADED" />
-    </Upgrade>
+    <!-- Launch condition to check for minimum OS: Windows 7/Server 2008R2. -->
+    <Condition Message="!(loc.MinimumOSVersionMessage)">
+      <![CDATA[Installed OR VersionNT >= 601]]>
```
btm: Are we bumping this for a feature we need or based on our supported platforms list?

lib/omnibus/packagers/msi.rb
chefsalim: Done

spec/unit/packagers/msi_spec.rb
chefsalim: Done

lib/omnibus/packagers/msi.rb
chefsalim: Done

resources/msi/source.wxs.erb
chefsalim: Not required for fast msi - it's a nice to have/best practice pattern that allows us to set our supported platforms and potentially make use of the latest installer features that come with the updated versions.

resources/msi/source.wxs.erb
chefsalim: See above

btm: Unless we know for a fact that Chef is completely broken on Server 2008, or even earlier, I don't want to prevent people from installing there. If it's working for those people, and they can take the benefit of future Chef releases, we want them to be able to. They're going to be much worse off if they keep using a versions of Chef with unfixed bugs.

I don't think any of our other packaging forces you to use only the versions of the platform listed in our supported platforms document.

If there's a feature we need for the installer that's only available on later Windows versions, we should weigh that trade-off directly.

chefsalim: Hmm, hang on - actually I don't know for sure that the CA will work as is for Windows prior to Window 7/Win 2008 R2.  Since it is targeting .Net 2.0 now, it may require additional work to either try and re-target for .Net 1.0, or have multiple versions, or build a bootstrapper for older versions. It's easy enough to change the minimum OS though either ways, so it things do work and we want to support earlier platforms, I would propose we leave in the minimum OS, but relax it to a version earlier to pick up Windows 2008 and Vista if we want to go that far back :)

btm: We should try the install on an earlier version of Windows, so we have a
clear answer in the release notes for the inevitable support ticket.

stevendanna:
:+1: on the refactoring on the Ruby side.

btm:
:+1:

btm:
@chef/omnibus-maintainers do we really need five thumbs ups on this (and omnibus-chef#540?) to move forward with it?

https://github.com/chef/omnibus/blob/master/MAINTAINERS.md#maintainers-1

yzl:
:+1:

PR #567, opened 2015-11-09
Update required ruby version to 2.0, added 2.2 to testing

btm:
@schisamo reports we're currently using 2.1.5 for our building.
It'd be nice to be able to use some of the new features of Ruby 2.0+.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 rvm:
   - 2.0.0
   - 2.1
+  - 2.2
 bundler_args: "--jobs 7 --without docs local"
 branches:
   only:
--- a/omnibus.gemspec
+++ b/omnibus.gemspec
@@ -13,7 +13,7 @@ Gem::Specification.new do |gem|
   gem.description    = gem.summary
   gem.homepage       = 'https://github.com/opscode/omnibus'
 
-  gem.required_ruby_version = '>= 1.9.1'
+  gem.required_ruby_version = '>= 2'
 
   gem.files = `git ls-files`.split($/)
   gem.bindir = 'bin'


.travis.yml
```diff
@@ -1,6 +1,7 @@
 rvm:
   - 2.0.0
```
schisamo: Should we change this to just `2.0`?

stevendanna:
:+1:

yzl:
:+1:

curiositycasualty:
:+1:

schisamo:
So if I'm being a SemVer pedant this feels like a change that would require a major version bump (e.g. `5.0.0`). BUT the pragmatist in me says most everyone was probably using Ruby 2+ already (the omnibus cookbook has installed this version for years). As my beard has greyed I've def become more of a pragmatist.

What does @chef/omnibus-maintainers think?

.travis.yml
sethvargo: Nah - it's only semver from 2.1+

danielsdeleo:
I could go either way. But for the sake of argument, what are the downsides to bumping to 5.0?

scotthain:
:+1: - I would vote to burn the version number and go to 5

stevendanna:
:+1: Moving to 5.0 is fine, but I'm with @schisamo that it probably isn't a big deal if we were hoping to get some bigger features in before a "major version" bump.

yzl:
:+1: on moving to 5.0

danielsdeleo:
We can just as easily bump to 6.0 later if we have substantive breaking changes we're planning for. Given the audience for this project, I think there's a lot less downside to bumping major more frequently compared to chef-client, for example.

schisamo:
I'm now convinced moving to `5.0` is the safest, least surprising, move. Thanks for the discussion y'all.

PR #571, opened 2015-11-20
Add a FIPS mode flag in omnibus.

ksubrama:
This will allow software definitions to pull in the right version of binaries and vary how it builds ruby accordingly.

### DIFF
--- a/lib/omnibus/config.rb
+++ b/lib/omnibus/config.rb
@@ -471,6 +471,12 @@ module Omnibus
       :x86
     end
 
+    # Should FIPS mode be enabled?  This will affect how ruby is build,
+    # how openssl is built and what libraries are vendored.
+    #
+    # @return [true, false]
+    default(:fips_mode, false)
+
     # --------------------------------------------------
     # @!endgroup
     #
--- a/lib/omnibus/sugarable.rb
+++ b/lib/omnibus/sugarable.rb
@@ -71,5 +71,10 @@ module Omnibus
     def windows_arch_i386?
       Config.windows_arch.to_sym == :x86
     end
+
+    # Returns if the global fips mode flag was enabled.
+    def fips_mode?
+      Config.fips_mode
+    end
   end
 end
--- a/spec/unit/sugarable_spec.rb
+++ b/spec/unit/sugarable_spec.rb
@@ -44,6 +44,8 @@ module Omnibus
         expect(klass).to be_method_defined(:windows?)
         expect(klass).to be_method_defined(:vagrant?)
         expect(klass).to be_method_defined(:_64_bit?)
+        expect(klass).to be_method_defined(:windows_arch_i386?)
+        expect(klass).to be_method_defined(:fips_mode?)
       end
 
       it 'makes the DSL methods available in the cleanroom' do
@@ -51,6 +53,8 @@ module Omnibus
           instance.evaluate <<-EOH.gsub(/^ {12}/, '')
             windows?
             vagrant?
+            windows_arch_i386?
+            fips_mode?
           EOH
         }.to_not raise_error
       end


jaym:
Most of this looks good. I'm not convinced that omnibus itself should know anything about fips. Is there something more general we could do here, such as provide options for software definitions or have software definitions that provide other software definitions? Or maybe some arguments for why we should have this `fips_mode` flag in omnibus.

ksubrama:
I thought of possibly making alternate software definitions for open-ssl that we could then override in a new copy of chef-client (chef-client-fips.rb or whatever) but that seemed like a lot of duplication.  If we eventually need to build push-jobs or chef-server or chef-dk or whatever with FIPS mode and they use omnibus, it made sense to have a flag in one place (like our 32-bit vs 64-bit windows architecture flag) and have the CI set that flag for the pipeline its running.

That's assuming we want a similar release cadence between non-fips and fips version of the software (maybe there might not be a difference eventually).  If we start releasing them and versioning them independently, then having 2 software definitions and just overriding the dependencies seems more reasonable.  There'll be some duplication - but no worse than angry-chef to chef (although perhaps just as annoying).

scotthain:
I'm :-1: on having omnibus 'know' anything about FIPS innately. I'd much prefer a flag or env variable that would add/remove the appropriate dependencies in the software definition. (Or some other solution, that's just the first thing I thought of)

jaym:
I'd be ok with an environment variable. While not great, it avoids any major replumbing of omnibus

schisamo:
So we need to separate the updates to enabling compiling more software from source on Windows from this FIPS flag change. I am in agreement Omnibus should know nothing about FIPS though.

schisamo:
And this probably goes without saying but we need test coverage for all these changes! ;)

ksubrama:
Yep - I think this'll be less scary if I split the two up.

ksubrama:
@schisamo @jaym @scotthain These should the core FIPS related changes.

scotthain:
I'm still :-1: on doing it this way. This basically means that omnibus the project has to care and pass on a config about fips, which it shouldn't ever care about.

I'd much rather see something in the `chef` software def that picks up an environment variable and conditionally adds different dependencies that already are fips enabled. (Or some other similar tactic). Ping me if what I'm saying isn't clear and we can chat @ksubrama

ksubrama:
I and @scotthain had a chat.  Ww were considering what possible options there were of not letting omnibus itself know or care about FIPS.  Ideally, if there was a way to let a project definition export a custom "we're doing FIPS guys" flag, that would let all the software definitions it depends on switch flags appropriately.  We didn't find an already existing way to do this.  The current options seem to be:

- have the project definition figure out if it's FIPS or not and then select an alternate software definition (like ruby vs ruby-windows) but the downside it lots of copy-pasta (ruby, ruby-fips, ruby-windows, ruby-windows-fips).

- have all the omnibus related software definitions and projects (client, server, ruby, openssl etc.) share an environment variable ENV['whatever'] and just access that directly that which feels a bit hacky as projects across different repos have a just remember what that ENV variable was rather than seeing it in one place.

- have omnibus support the ability of projects/software to pass/receive custom flags other than just source/version etc.  This would be ideal but seems like a new DSL feature to implement and test and I'd need to sit down with y'all to discuss how it should be architected and how it would evolve.

Currently it seems like the solaris mapfile flag in config.rb suffers from the same issue.  I'll take any ideas we have to solve that problem.

ksubrama:
Actually - @scotthain apparently did solve the problem in one of his branches. https://github.com/chef/omnibus-software/blob/shain/woo_fips/config/software/openssl.rb#L31

I'll look at that now

ksubrama:
I withdraw this PR.  @scotthain 's crever reuse of the overrides hash is exactly what I needed.

PR #572, opened 2015-11-23
Enable building from source on windows

ksubrama:
This brings a lot of feature parity between building dlls/exes from source using mingw on windows (when compared to building from source on linux).


lib/omnibus/builder.rb
```diff
@@ -105,27 +108,101 @@ def command(command, options = {})
...
+
+      make_cmd = ([make] + args).join(' ').strip
+      if windows?
+        # The msys make that we deploy requires that we first invoke a shell.
+        # Fire up bash.
+        # TODO: Eventually, have make search through "build time dependencies"
+        # for bash/make instead.
+        bash_bin = embedded_msys_bin('bash.exe')
+        # TODO: I need to powershell/cmd escape command.strip.... :( No.
+        # We're going to use ' and hope for the best.
```
schisamo: You might be able to take advantage of the Ruby Standard Library `Shellwords.escape`:
http://ruby-doc.org/stdlib-2.0.0/libdoc/shellwords/rdoc/Shellwords.html#method-c-escape

schisamo:
@ksubrama Over all this is a great start. Should we be concerned about the case where people might want to perform their Windows compilations with MSVC?

A few other comments:

* We obviously need tests passing
* Be sure to squash things down before a final review

danielsdeleo:
Looks good overall. Needs passing tests, of course.

ksubrama:
@schisamo Fixed the shell-escaping you wanted and the tests now pass.  I'm squashing now.

PR #573, opened 2015-11-30
Default to UTF-8 external encoding globally.

edolnx:
Many omnibus projects will not build properly without UTF-8
encoding. This sets the default external encoding to UTF-8
via ruby, which works even when the platform's locale is C or
some other non-UTF-8 encoding. UTF-8 is backwards compatible with
C, POSIX, and ASCII so there should be no issues. The default
can be overridden the per-project omnibus.rb file still.

@chef/omnibus-maintainers

### DIFF
--- a/bin/omnibus
+++ b/bin/omnibus
@@ -8,4 +8,9 @@ $:.push File.expand_path('../../lib', __FILE__)
 $stdout.sync = true
 
 require 'omnibus/cli'
+
+# Some platforms do not have a UTF-8 locale, so we need to enforce one
+# or else the cacert chain will break among other things
+Encoding.default_external = Encoding::UTF_8
+
 Omnibus::CLI::Runner.new(ARGV.dup).execute!


lamont-granquist:
:+1: 

I doubt anyone wants a different default encoding for their entire project.

On a case-by-case basis I could see windows files maybe needing to be read and written in UTF-16, but changing the global default external encoding for the whole omnibus run from UTF-8 to UTF-16 will be an incredibly poor solution to that problem--if it occurs.  That should be handled in the software defn that needs to deploy those particular files in that encoding and/or just via magic comments in the header of the file in question to change the encoding of that specific file.

lamont-granquist:
And if we do go down the road of actually finding a bug where the correct solution is not to set the encoding on a specific file but an entire project needs a different encoding then users can still as @edolnx says, override this with code in omnibus.rb, or we can then spend the time to make this a config option once we see the bug that actually requires it (and until then YAGNI).

scotthain:
:+1:

schisamo:
:+1:

PR #499, opened 2015-06-23
remove files with colons in name

curiositycasualty:
Replete with inline commentary. I've seen "::" filenames from the git and perl manual page files thus far.

PR #575, opened 2015-12-02
Rename and restore invalid file/paths on AIX

curiositycasualty:
After reviewing https://www-01.ibm.com/support/knowledgecenter/ssw_aix_61/com.ibm.aix.genprogc/pkging_sw4_install.htm this seems like a more complete solution than https://github.com/chef/omnibus/pull/499

Essentially, this creates a BFF package 'config' script consisting of just a bunch of `mv` commands that rename files which had invalid characters during packaging. This 'config' script is run after both the actually restoration of the files and the sysck check have occurred.

/cc @chef/omnibus-maintainers

### DIFF
--- a/lib/omnibus/packagers/bff.rb
+++ b/lib/omnibus/packagers/bff.rb
@@ -21,6 +21,8 @@ module Omnibus
       # Default Omnibus naming
       preinst:  'Pre-installation Script',
       postinst: 'Post-installation Script',
+      config:   'Configuration Script',
+      unconfig: 'Unconfiguration Script',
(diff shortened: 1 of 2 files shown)


schisamo:
Looks good overall. Can we add some unit test coverage to [spec/unit/packagers/bff_spec.rb](https://github.com/chef/omnibus/blob/master/spec/unit/packagers/bff_spec.rb) for this new logic?

scotthain:
:+1: modulo tests, if possible.

scotthain:
Still :+1: pending travis run :)

PR #576, opened 2015-12-03
Properly decompress `.tar.lzma` files

schisamo:
This builds on the `.tar.lzma` support added in chef/omnibus#572. We should properly decompress the file before extracting. This follows the pattern followed by other compressed files like `.tar.gz` and `.tar.bz2`.

/cc @chef/omnibus-maintainers @ksubrama


yzl:
:+1:

scotthain:
:+1:

curiositycasualty:
:+1: oof

danielsdeleo:
:+1:

PR #578, opened 2015-12-04
Use 7z on windows instead of tar.exe

jaym:
tar.exe has issues when tar files have read-only files.
When a read only file exists on disk, tar will be unable
to overwrite the file.

cc @chef/omnibus-maintainers @chefsalim

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ branches:
     - 2.0-stable
     - 3.0-stable
     - master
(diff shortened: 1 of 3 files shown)


schisamo:
:+1:

yzl:
:+1:

curiositycasualty:
:+1:

ksubrama:
:+1:

PR #552, opened 2015-09-29
Issue#509: Reset index and working tree to original clone

hugosato:
Original the problem arrises when software uses GitFetcher and build fails **after** a patch is applied. Currently, in the next build attempt, only `git clean -fdx` is called in #{project_dir}, as result any modification to a tracked file is left un-reverted in the working tree. 

This can become an automated build blocker if there is an intermittent build failure. E.g.: timeout while running `bundle install` in Builder after a patch is applied. 

Let me know what tests to run or update.

PR #581, opened 2015-12-10
Simplify behavior of all fetchers.  Make net_fetcher always put files in project_dir.

### LINKED ISSUE #509: Reapplying a patch results in an error
I have a patch for a Makefile to fix a lib path, auto configure did not pick up the dependency path correctly.  It works just fine on the a fresh run, but if the cache is dirty and the software needs to recompile I get an error.  The error is related to the patch being applied previously.  I am not sure if I am just using patch incorrectly.  Is there a way to write a guard for the patch method?

The error I get is:
```
    patching file src/Makefile.am
Reversed (or previously applied) patch detected!  Assume -R? [n]
Apply anyway? [n]
Skipping patch.
1 out of 1 hunk ignored -- saving rejects to file src/Makefile.am.rej
```

It exits with error code 1 which Omnibus picks up as a failure.  The workaround is to delete the src directory for the specific software so the original files are extracted again.

ksubrama:
As part of simplifying this, I had noticed that fetcher.clean's semantics were confusing to me. The boolean return value seems to roughly mean "did you have to clean it up?" but the method was also unconditionally called for exactly one place - Omnibus::Software, and it was only going to do so after fetcher.fetch was called. So I simplified a bunch of the logic. Now, all the clean methods assume that "fetch" is called before it is (assuming that a fetch was required). clean only performs any extra work necessary to tidy up the project_dir assuming that some stale content may be present.

The tests have been reflected accordingly. This cleanup also fixes #509 by subsuming the fix in #552. I just use git checkout -f instead of git reset --hard.

Reviewing them as 4 separate but similarly themed commits should make things easier.


lib/omnibus/fetchers/net_fetcher.rb
```diff
@@ -125,6 +125,11 @@ def checksum
       source[digest_type]
     end
 
+    # Set to false to extract archives into the root source_dir (default).
+    # Set to true to always extract underneath project_dir
+    #
+    attr_accessor :isolate_extraction
```
schisamo: Why do we need this extra bit of configuration? Shouldn't we always isolate extraction? It seems that behavior would be fully backward compatible per any public APIs/DSL we current expose.

ksubrama: The intent of the flag is to let fetcher know whether it should extract to Config.source_dir or to project_dir since the previous meaning of "relative_path" was to provide some path from Config.source_dir.  Maybe I could call the flag "legacy_behavior" or something like that?

scotthain: if this does change that behavior, won't this essentially break backwards compat? (Not that I'm against that, just need to make sure we version and discuss correctly)

jaym: I'm having the same thoughts as @schisamo 
I think this can be implemented in a fully backwards compatible way without an extra config option (I could be wrong, I dont know all the omnibus things):

Everything should start extracting into a folder exclusive for that package. Whenever builder steps are executed, they should be with cwd `source_dir/package_path/relative_path`. For this windows stuff, we dont specify relative path, and it's just `source_dir/package_path`. `source_dir` is something liek C:\omnibus-src\

ksubrama: I guess it depends on what you mean by "backwards compatible".  See https://github.com/chef/omnibus-software/blob/master/config/software/libzmq4x-windows.rb  It assumes that the extracted zip is going to be right under source_dir.  See https://github.com/chef/omnibus/blob/master/lib/omnibus/builder.rb#L355 .  It assumes that the gem you're building is going to be right under source_dir.  I'm sure there are other such implicit assumptions in our CI and in other omnibus definitions.  Now I don't mind breaking that and making people fix their code... but I was wary of fighting too many yaks at once.  If y'all are cool with that, then, indeed, I can make this way simpler.  No need for extract_path.  Every fetcher will always unconditionally extract to a source_dir/package_path and the software DSL will take care of correctly setting project_path to source_dir/package_path/relative_path.

schisamo: @ksubrama I believe that example is bad since the software definition could have been written in such a way to use `relative_path`

So this:

```
relative_path 'libzmq4x-windows`

build do
  copy "#{project_dir}/bin/*", "#{install_dir}/embedded/bin"
  copy "#{project_dir}/include/*", "#{install_dir}/embedded/include"
  copy "#{project_dir}/lib/*", "#{install_dir}/embedded/lib"

  # Ensure the main DLL is available under a well known name.
  copy "#{project_dir}/bin/libzmq-mt-4_0_6.dll", "#{install_dir}/embedded/bin/libzmq.dll"
end
```
instead of
```
build do
  tmpdir = File.join(Omnibus::Config.source_dir, "libzmq4x-windows")

  copy "#{tmpdir}/bin/*", "#{install_dir}/embedded/bin"
  copy "#{tmpdir}/include/*", "#{install_dir}/embedded/include"
  copy "#{tmpdir}/lib/*", "#{install_dir}/embedded/lib"

  # Ensure the main DLL is available under a well known name.
  copy "#{tmpdir}/bin/libzmq-mt-4_0_6.dll", "#{install_dir}/embedded/bin/libzmq.dll"
end
```

Regardless I don't think we should create new config options for 1 or 2 cases we can easily fix.

jaym: lol, stupid past me. it should be using `relative_path`

schisamo: I personally think we should nuke this config option and do some testing in the context of the Chef project. We are changing internal behavior IMO.

lib/omnibus/fetcher.rb
```diff
@@ -55,7 +55,17 @@ class Fetcher
     attr_reader :described_version
 
     #
-    # The path where extracted software should live.
+    # The path where fetched software should live.
+    #
+    # Only files under this directory are modified. If the source to fetch
+    # is a directory, it is staged rooted here. If it's a file, it's copied
+    # underneath this directory. If it's a tarball, it's extracted here. If
+    # it's a repo, it's checkout is rooted here. You get the idea.
```
stevendanna: 'it's checkout' -> 'its checkout'

lib/omnibus/fetchers/net_fetcher.rb
ksubrama: I'm convinced.

The extra piece of info that swayed me was realizing that the comments below for line 218 where the code was mistakenly checking if project_dir was a directory instead of downloaded_file (project_dir is always going to be a directory) don't even matter.  We were always expecting to copy stuff into a directory and other software definitions like cacert.rb were written to work around this (which I found out while attempting a test build).  So this was always a bit wonky and not doing the right thing anyway.

I'll go clean up this PR with a chainsaw now.

ksubrama:
Ok - @schisamo @scotthain @jaym.  Round 2.

As part of simplifying this, I had noticed that fetcher.clean's semantics were confusing to me.  The boolean return value seems to roughly mean "did you have to clean it up?" but the method was also unconditionally called for exactly one place - Omnibus::Software, and it was only going to do so after fetcher.fetch was called.  So I simplified a bunch of the logic.  Now, all the clean methods assume that "fetch" is called before it is (assuming that a fetch was required).  clean only performs any extra work necessary to tidy up the project_dir assuming that some stale content may be present.

The tests have been reflected accordingly.  This cleanup also fixes #509 by subsuming the fix in #552.  I just use git checkout -f instead of git reset --hard.

jaym:
A bit of a nit, but I would love to see the commit message for 13ad79db55a3bb5b7cc8376fcd5705346d3c84c7 include the reason for the change. It's not clear from the title of this PR or the contents of the commit message, which are the 2 places I tend to look when I want to know why something was done

lib/omnibus/fetchers/git_fetcher.rb
```diff
@@ -169,6 +167,17 @@ def current_revision
...
 
     #
+    # Check if the current clone has the requested commit id.
+    #
+    # @return [true, false]
+    #
+    def contains_revision?(rev=resolved_version)
+      git("cat-file -t #{resolved_version}").stdout.strip == 'commit'
+    rescue CommandFailed
+      false
```
jaym: Should we need to log something here?

jaym:
:+1:

ksubrama:
@jaym I have better commit messages for you :)

Please ignore the latest omnibus failure.  Here's it passing using bundler 1.10.6.  Bundler 1.11.0 has a bug where it won't fetch certain gems on windows because of a bug.

https://ci.appveyor.com/project/chef/omnibus/build/189

scotthain:
Looks pretty solid.

lib/omnibus/software.rb
```diff
@@ -358,12 +363,49 @@ def relative_path(val = NULL)
...
+    # @return [String] the full absolute path to the package root directory.
+    #
+    def package_dir(val = NULL)
+      if null?(val)
+        @package_dir || File.expand_path("#{Config.source_dir}/#{name}")
+      else
+        @package_dir = File.expand_path(val, Config.source_dir)
+      end
+    end
+    expose :package_dir
```
schisamo: Why are we exposing this in the public DSL?

PR #584, opened 2015-12-14
Rename package_dir to fetch_dir and is fix its behavior for back-compat

ksubrama:
Allows for better back-compat for changes introduced in #581

### DIFF
--- a/lib/omnibus/software.rb
+++ b/lib/omnibus/software.rb
@@ -338,11 +338,11 @@ module Omnibus
     expose :whitelist_file
 
     #
-    # The path relative to package_dir where relevant project files are
+    # The path relative to fetch_dir where relevant project files are
     # stored. This applies to all sources.
     #
     # Any command executed in the build step are run after cwd-ing into
-    # this path. The default is to stay at the top level of package_dir
+    # this path. The default is to stay at the top level of fetch_dir
     # where the source tar-ball/git repo/file/directory has been staged.
     #
     # @example
@@ -350,6 +350,12 @@ module Omnibus
     #
     # @param [String] val
     #   the relative path inside the source directory. default: '.'
+    #
+    # Due to back-compat reasons, relative_path works completely
+    # differently for anything other than tar-balls/archives. In those
+    # situations, the source is checked out rooted at relative_path
+    # instead 'cause reasons.
+    # TODO: Fix this in omnibus 6.
     #
     # @return [String]
     #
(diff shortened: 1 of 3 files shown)


lib/omnibus/software.rb
```diff
@@ -351,6 +351,12 @@ def whitelist_file(file)
     # @param [String] val
     #   the relative path inside the source directory. default: '.'
     #
+    # Due to back-compat reasons, relative_path works completely
+    # differently for anything other than tar-balls/archives. In those
+    # situations, the source is checked out rooted at relative_path
+    # instead 'cause reasons.
+    # TODO: Fix this in omnibus 5.
```
schisamo: We are in Omnibus 5.

lib/omnibus/software.rb
```diff
@@ -767,6 +746,19 @@ def overridden?
...
+    #
+    # Path to where any source is extracted to.
+    #
+    # Files in a source directory are staged underneath here. Files from
+    # a url are fetched and extracted here. Look outside this directory
+    # at your own peril.
+    #
+    # @return [String] the full absolute path to the package root directory.
+    #
+    def package_dir
```
schisamo: Do we still think this is the best name for this function? In omnibus packages (and packagers) are the output of a build. I don't see how this is related to packages.

scotthain:
:+1:

lib/omnibus/software.rb
ksubrama: I remember you mentioning fetch_dir.  I'm ok with that if you are.  I can s/package_dir/fetch_dir/g.

curiositycasualty: `extract_dir` might be better, but I'd much rather see that come in a later PR.

yzl: Maybe `source_staging_dir`?

yzl:
:+1: modulo discussion of method name.

lib/omnibus/software.rb
ksubrama: Actually seeing as how package_dir is used in the packages, I would be a terrible person to disambiguate project_dir with "the directory this software def is built in" only to then confuse it with the packager output.  I've updated the PR by calling it fetch_dir instead.  It might be a bit weird inside the git_fetcher (since its "project_dir" might be the "fetch_dir" where neither project nor fetch have anything to do with git), I think in isolation, each file makes fairly good sense with that name.  Once we have the options/ability to lose the back-compat features, it should be a whole lot less weird.

ksubrama: Woah eventual consistency model....

Ok..  so would someone prefer extract_dir/source_staging_dir instead?  It's still the same regex and I honestly want to pick the name y'all like best since you're the ones who have to deal with it and I've already spend 4-5 days being bewildered with the code and don't want to inflict the same on you.

schisamo:
:+1:

PR #586, opened 2015-12-16
More windows source building features/fixes.

ksubrama:


lib/omnibus/builder.rb
```diff
@@ -765,10 +764,11 @@ def shellout!(command_string, options = {})
...
         end
-        # Mixlib will handle escaping characters for cmd but we need to escape
-        # the command for bash itself. Since we're using Shellwords to escape
-        # the command, we should not quote the string at all.
-        command_string = "\"#{bash_bin}\" -c #{Shellwords.escape(command_string)}"
+        # Mixlib will handle escaping characters for cmd but our command might
+        # contain '. For now, assume that won't happen because I don't know
+        # whether this command is going to be played via cmd or through
+        # ProcessCreate.
+        command_string = "\"#{bash_bin}\" -c \'#{command_string}\'"
```
ksubrama: I couldn't use Shellwords.escape as that caused bash to interpret the entire string as one command (i.e. the name of the binary).  Shellwords.escape would be appropriate here if we were using bash to launch this.

lib/omnibus/software.rb
```diff
@@ -585,11 +585,15 @@ def with_standard_compiler_flags(env = {}, opts = {})
...
     #
     # @param [Hash] env
+    # @param [Hash] opts
+    #   :msys => true  add the embedded msys path if building on windows.
     #
     # @return [Hash]
     #
-    def with_embedded_path(env = {})
-      path_value = prepend_path("#{install_dir}/bin", "#{install_dir}/embedded/bin")
+    def with_embedded_path(env = {}, opts = {})
```
ksubrama: Needed a way to add msys paths after embedded/bin but before other stuff like git (which also has bash/less/etc.).

lib/omnibus/software.rb
```diff
@@ -585,11 +585,15 @@ def with_standard_compiler_flags(env = {}, opts = {})
...
+      paths = ["#{install_dir}/bin", "#{install_dir}/embedded/bin"]
+      paths << "#{install_dir}/embedded/msys/1.0/bin" if opts[:msys] && windows?
```
schisamo: We should add some test coverage to:
https://github.com/chef/omnibus/blob/master/spec/unit/software_spec.rb#L283-L316

ksubrama: Done!

schisamo:
:+1: 

/cc @chef/omnibus-maintainers

scotthain:
Not up on my windows enough to know if the invocations are correct but the good looks good to me :+1:

PR #587, opened 2015-12-16
Update logic for finding default number of workers

edolnx:
Inside a Solaris Zone, solaris returns the incorrect number of
CPUs (0). This will now detect that and return the default of 3
instead. This should be suitable for any environment.

### DIFF
--- a/lib/omnibus/config.rb
+++ b/lib/omnibus/config.rb
@@ -446,7 +446,7 @@ module Omnibus
     #
     # For example:
     #
-    #     /PATH/files/my_map_file 
+    #     /PATH/files/my_map_file
     #
     # @return [String, nil]
     default(:solaris_linker_mapfile, "files/mapfiles/solaris")
@@ -498,12 +498,17 @@ module Omnibus
 
     # The number of worker threads for make. If this is not set
     # explicitly in config, it will attempt to determine via Ohai in
-    # the builder, and failing that will default to 3
+    # the builder, and failing that will default to 3. This will
+    # also default to 3 if the count is 0 (like in a Solaris zone)
     #
     # @return [Integer]
     default(:workers) do
-      if Ohai['cpu'] && Ohai['cpu']['total']
-        Ohai['cpu']['total'].to_i + 1
+      count = 0
+      if Ohai['cpu'] && Ohai['cpu'].key?(:total)
+        count = Ohai['cpu']['total'].to_i
+      end
+      if count > 0
+        count + 1
       else
         3
       end


edolnx:
cc @chef/omnibus-maintainers

lib/omnibus/config.rb
```diff
@@ -498,12 +498,17 @@ def reset!
...
-    # the builder, and failing that will default to 3
+    # the builder, and failing that will default to 3. This will
+    # also default to 3 if the count is 0 (like in a Solaris zone)
     #
     # @return [Integer]
     default(:workers) do
-      if Ohai['cpu'] && Ohai['cpu']['total']
-        Ohai['cpu']['total'].to_i + 1
+      count = 0
+      if Ohai['cpu'] && Ohai['cpu'].key?(:total)
```
lamont-granquist: if its a Mash this should probably be 'total' since that's consistent with the rest of the style here, and (our) Mashes convert to strings under the covers anyway.

lamont-granquist: Also, though, how is 'Ohai['cpu']['total']' getting set to a value that isn't a string-of-an-integer?  If 'nil' is showing up here that's an ohai bug.

edolnx:
There were previous ohai bugs on some platforms that was causing that to yield non-numbers. This is mostly to squash those. Also, some platforms do not have a total key in the cpu module, that's the reasoning for this.

lamont-granquist:
So isn't this code actually worse?

The previous code protected against a trainwreck and avoided nil values because `if Ohai['cpu'] && Ohai['cpu']['total']`would be false and not wreck if either the `cpu` or the `total` keys were missing, and it would also be falsey in the case that the `total` key had a value of nil.

By only testing for the presence of the key `total` now that is true for a nil value and you fall through and you should get a NoMethodError on NilClass when you try to `#to_i` that.

The introduction of the count variable also is just complicating the code without adding anything.

lamont-granquist:
Oh, I see you care about `"".to_i` being '0'.

```ruby
default(:workers) do
  cores =
    if Ohai['cpu'] && Ohai['cpu']['total']
      # this may return '0' if the string is not a number
      Ohai['cpu']['total'].to_i
    else
      2
    end

  [ 3, cores + 1 ].max
end
```

The only issue with that code is that a box that really has 1 core will get 3 procs instead of 2, but will anyone care that much?

You could make the conditional `Ohai['cpu'] && Ohai['cpu']['total'] &&  Ohai['cpu']['total'].respond_to?(:to_i)` i guess for maximum bug proofing...

edolnx:
It is adding something: when running inside a solaris zone the CPU count is 0. This is false, and now we say "if zero, switch to the default". Previous code would return 1, or sometimes 254 (I don't know how that happened, but everything was forced to 10 workers because of it). This is trying to make it more sane and predictable. If you don't agree, that's fine we can ignore it. I need to put other changes to force workers to 10 on our Solaris build code anyway, and then limit workers to no more than 10 in Ruby.

lamont-granquist:
^ i'd prefer you used my code there.  pretty sure your code will trainwreck on Ohai[:cpu][:total] being set to nil and its harder to follow your logic.

edolnx:
I don't see how that will do the same thing as my code currently. Currently, Ohai['cpu']['total'] in a Solaris zone will return 0. Your code will return a worker count of 1 in this case, I wanted the default of 3 returned instead. This will also return a value of 3 on single core systems, which is less than ideal. I'm just going to leave the existing behavior as is and close this PR. The problem will need to be solved elsewhere.

lamont-granquist:
Also pretty clearly this needs specs tests for the cases:

- Ohai['cpu'] does not exist
- Ohai['cpu']['total'] does not exist
- `Ohai['cpu']['total'] = nil`
- `Ohai['cpu']['total'] = "4"`
- `Ohai['cpu']['total'] = 4`
- `Ohai['cpu']['total'] = ""`
- `Ohai['cpu']['total'] = "randomgarbage"`

lamont-granquist:
My code actually always returns a minimum value of '3' in all cases.  So on a solaris zone where cores is 0 you get 3.   That does mean on a 1 core box you get 3. 

This should get everything right:

```ruby
default(:workers) do
  cores =
    if Ohai['cpu'] && Ohai['cpu']['total'] && Ohai['cpu']['total'].respond_to?(:to_i)
      # this may return '0' if the string is not a number
      Ohai['cpu']['total'].to_i
    else
      2
    end

  cores = 2 if cores == 0
  cores + 1
end
```

PR #588, opened 2015-12-16
Fix typo when generating log message.

ksubrama:


schisamo:
Why didn't the tests catch this? We should add coverage now so it doesn't happen again.

ksubrama:
I'm not sure what the best way to add coverage is.  We set Omnibus.logger.level = :nothing in spec_helper and the type I made was inside an interpolated string sent to the logger.  So it didn't blow up because the log statement was never evaluated.

schisamo:
@ksubrama Line 177 isn't inside a log statement right?

ksubrama:
177 wasn't syntactically incorrect.  We never set that parameter to anything but resolved_version - so none of the functional tests failed for that.

schisamo:
@ksubrama Looks good. Let's just remove 839efbc33e0ab3aa31ea29106907598337c42012 and squash this down.

PR #591, opened 2015-12-18
Use -O2 when building with standard compiler flags

stevendanna:
Since we are setting CFLAGS explicitly, we override any Makefile
defaults, which will often be -O2.

### DIFF
--- a/lib/omnibus/software.rb
+++ b/lib/omnibus/software.rb
@@ -492,7 +492,7 @@ module Omnibus
         when "mac_os_x"
           {
             "LDFLAGS" => "-L#{install_dir}/embedded/lib",
-            "CFLAGS" => "-I#{install_dir}/embedded/include",
+            "CFLAGS" => "-I#{install_dir}/embedded/include -O2",
           }
         when "solaris2"
           {
@@ -505,7 +505,7 @@ module Omnibus
         when "freebsd"
           freebsd_flags = {
             "LDFLAGS" => "-L#{install_dir}/embedded/lib",
-            "CFLAGS" => "-I#{install_dir}/embedded/include",
+            "CFLAGS" => "-I#{install_dir}/embedded/include -O2",
           }
           # Clang became the default compiler in FreeBSD 10+
           if Ohai['os_version'].to_i >= 1000024
@@ -532,7 +532,7 @@ module Omnibus
         else
           {
             "LDFLAGS" => "-Wl,-rpath,#{install_dir}/embedded/lib -L#{install_dir}/embedded/lib",
-            "CFLAGS" => "-I#{install_dir}/embedded/include",
+            "CFLAGS" => "-I#{install_dir}/embedded/include -O2",
           }
         end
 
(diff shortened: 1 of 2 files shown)


stevendanna:
Chef appears to build without issue given this change: http://manhattan.ci.chef.co/job/chef-12-trigger-ad_hoc/8/downstreambuildview/

schisamo:
Looks like we need to update some tests @stevendanna 

/cc @chef/omnibus-maintainers

stevendanna:
@schisamo Indeed.  Before I do that, any thoughts on whether we should also specify `-g`?

schisamo:
@stevendanna I have no strong feelings, what say you @lamont-granquist ?

scotthain:
I don't think `-g` is necessary for general compilation. It's great to add in the corner case of something breaking, but you generally don't need debug info...

stevendanna:
Works for me.  Updated tests incoming.

danielsdeleo:
:+1: once there's passing tests

lamont-granquist:
I've never needed to recompile with -g yet, and I don't think anyone else has, so YAGNI...

I think -g may also turn off some optimizations, particularly on non-gcc compilers...

lamont-granquist:
@stevendanna i think it should be fine on slowaris and windoze as well...

stevendanna:
@scotthain Advised against solaris.  I defer to others on windows, happy to add them though.

lamont-granquist:
slowaris is just compiled with gcc though isn't it?
 
same with windoze...

and particularly with windoze, that platform is extremely slow, and this might help speed it up a bit for users, we have a perpetually open workflow/ease-of-use bug on windows perf needing improvement...

scotthain:
@lamont-granquist I'm willing to try it on slowlaris if you think it might help with optimization. I vaguely recall some issues since it's stuck at GCC 3.4.3 and the rest are running 4.x

stevendanna:
will add them to both and run them through manhattan.  Is there anything else I need to test other than a manhattan run?

lamont-granquist:
The way I look at it, if you look back in time far enough there's always some issue with -O2 screwing up optimizations.  Even if you were on bleeding edge gcc, you just don't know what -O2 bugs it has.  The majority of the time, though, people don't get bit by the -O2 optimizer.  Unless you know there's a specific optimizer bug with that version of GCC and ruby, its probably as safe to use -O2 as it is on new gcc.

lamont-granquist:
I'd say just make sure that this doesn't sneak into 12.6.0 and get released before christmas.  Make sure that it gets a full cycle of manhattan runs and only winds up in 12.7.0 in Jan.

What I'd be most worried about would be segfaults that happen intermittently, so we'd want 5-10 runs through manhattan before we start to trust it...

scotthain:
I'm +1 on what @lamont-granquist says for timing and vetting.

lamont-granquist:
@stevendanna can we click merge?

yzl:
:+1:

lamont-granquist:
well, tests pass...   we've never had coverage for these...

PR #592, opened 2015-12-22
Initial WIP for automatically updating config.guess via DSL call

scotthain:
:construction: 

@lamont-granquist @edolnx @curiositycasualty  @chef/omnibus-maintainers 

Based on Lamont's idea here: https://github.com/chef/omnibus-software/pull/511

More to get some kind of idea if I'm going in the right direction, and see if anyone wants to work with me on it.

What works:
- `update_config_guess` - puts stuff into the root directory of the current software def's src folder
- `update_config_guess('build-aux')` - will put it into `src-folder/build-aux` in this example

What doesn't work:
- adding a revision/timestamp for 2nd parameter. I didn't have time to dig into the GitFetcher to pass the right info

I didn't test the `config_guess_only` etc logic.

Some concerns I have about this in general that I'm not sure how to fix:
- The timing of when this happens is weird. Ideally it would fit into a `BuildCommand` step and go after the previous `BuildCommand`, but I couldn't figure out how to slot it in there and still utilize the GitFetcher sort of 'on the fly'. I'm sure someone smarter than I can figure this out.
- My Ruby code style is not super great so any refactors for clarity and "don't abuse the Ruby like this"
- Tests! TESTS! I started writing some then my head started to hurt due to the "when things run" issue. (I probably was doing it wrong.)

### DIFF
--- a/lib/omnibus/builder.rb
+++ b/lib/omnibus/builder.rb
@@ -635,6 +635,37 @@ module Omnibus
     end
     expose :sync
 
+
+    def update_config_guess(target = '', version = 'master', options = {})
+      destination = File.join(software.project_dir, target)
+
+      c_g_data = {
+        :locked_version => version,
+        :locked_source => {
+          :git=>"http://git.savannah.gnu.org/r/config.git"
+        },
+        :source_type => 'git',
+        :described_version => version
+      }
+
+      config_guess_manifest = ManifestEntry.new('config_guess', c_g_data)
+
+      config_guess_dir = File.join(Config.source_dir, "config_guess-#{version}")
+      config_guess_fetcher = GitFetcher.new(config_guess_manifest, config_guess_dir, destination)
+
+      config_guess_fetcher.fetch
+
+      if options[:config_guess_only] == true
+        copy("#{config_guess_dir}/config.guess", destination)
+      elsif options[:config_sub_only] == true
+        copy("#{config_guess_dir}/config.sub", destination)
+      else
+        copy("#{config_guess_dir}/config.guess", destination)
+        copy("#{config_guess_dir}/config.sub", destination)
+      end
+    end
+    expose :update_config_guess
+
     #
     # @!endgroup
     # --------------------------------------------------


lamont-granquist:
@edolnx here's the PR

lib/omnibus/builder.rb
```diff
@@ -635,6 +635,37 @@ def sync(source, destination, options = {})
...
+      config_guess_fetcher.fetch
+
+      if options[:config_guess_only] == true
+        copy("#{config_guess_dir}/config.guess", destination)
+      elsif options[:config_sub_only] == true
+        copy("#{config_guess_dir}/config.sub", destination)
+      else
+        copy("#{config_guess_dir}/config.guess", destination)
+        copy("#{config_guess_dir}/config.sub", destination)
+      end
```
lamont-granquist: can be replaced with:

```ruby
copy("#{config_guess_dir}/config.guess", destination) unless options[:config_sub_only]
copy("#{config_guess_dir}/config.sub", destination) unless options[:config_guess_only]
```

you might consider having `options[:install]` take `:all`, `:config_guess`, `:config_sub` so you don't have to worry about thinking about someone setting both config_guess_only and config_sub_only to true...

lamont-granquist:
i don't think we need to care about the revision/timestamp for now.

it needs to happen early and its not really a 'build step', so i think your 'timing issue' is probably correct behavior?

lamont-granquist:
:+1: in general

lib/omnibus/builder.rb
```diff
@@ -635,6 +635,37 @@ def sync(source, destination, options = {})
...
+          :git=>"http://git.savannah.gnu.org/r/config.git"
+        },
+        :source_type => 'git',
+        :described_version => version
+      }
+
+      config_guess_manifest = ManifestEntry.new('config_guess', c_g_data)
+
+      config_guess_dir = File.join(Config.source_dir, "config_guess-#{version}")
+      config_guess_fetcher = GitFetcher.new(config_guess_manifest, config_guess_dir, destination)
```
schisamo: So this magic scares me. I believe it would best to treat the `update_config_guess ` DSL function like we do `bundle`:
https://github.com/chef/omnibus/blob/master/lib/omnibus/builder.rb#L315-L318

Basically a project that wants to use the `bundle` DSL function should ensure Bundler is part of the project dependency tree. So in this case the assumption would be a `config` software definitoin is a dep of any software def that will call `update_config_guess ` in their build block.

schisamo:
We need full test coverage for this new DSL function also.

lib/omnibus/builder.rb
scotthain: I put this magic in this location as a POC because there was some discussion here about the best way to do it: https://github.com/chef/omnibus-software/pull/511

schisamo: So I agree we need the `update_config_guess` DSL function in Omnibus core but the fetching of a dependency (even just build time) up in the bowls of Omnibus is too hacky. I'm suggesting a hybrid of chef/omnibus-software#511 and this PR.

lamont-granquist: I like the analogy to the `bundle` DSL function.

scotthain:
updated with https://github.com/chef/omnibus/pull/632

PR #597, opened 2016-01-04
ChefDK is getting cached into a subfolder (chef-dk/chefdk) so to appbundle it we need to point to the correct directory

tyler-ball:
This is to fix ChefDK builds on Manhattan

\cc @jaym @danielsdeleo @schisamo

### DIFF
--- a/lib/omnibus/builder.rb
+++ b/lib/omnibus/builder.rb
@@ -340,18 +340,26 @@ module Omnibus
     # @example
     #   appbundle 'chef'
     #
+    # @param software_name [String]
+    #  The omnibus software definition name that you want to appbundle.  We
+    #  assume that this software definition is a gem that already has `bundle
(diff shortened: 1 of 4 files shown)


jaym:
:+1:

tyler-ball:
Per discussion with @jaym I added another commit to this PR.  One of the things we saw when troubleshooting this is that chefdk was cached in `/var/cache/omnibus/src/chefdk/chef-dk` and chef was cached into `/var/cache/omnibus/src/chef`.  We want to change that and ensure all software packages are cached into `.../src/<software definition name>/<package>`.

The current change fixes this for Chef, Berkshelf, etc.  It _does not_ change `chef-provisioning-aws` - we don't know why.  Our guess is because `relative_path` is not defined on the `chef-provisioning-aws` software definition.  @ksubrama 

But this PR moves the needle closer!

danielsdeleo:
:+1:

PR #599, opened 2016-01-06
Reverting https://github.com/chef/omnibus/pull/583 to try and fix ChefDK build errors

tyler-ball:
\cc @jaym @danielsdeleo @schisamo @ksubrama 

This fixes the builds because (we think) the makefile for libgecode isn't appropriate leveraging the `ARFLAGS` environment variable.  This causes the ar command to lose its operation (add, replace, etc.).

### DIFF
--- a/lib/omnibus/software.rb
+++ b/lib/omnibus/software.rb
@@ -525,9 +525,7 @@ module Omnibus
           end
           {
             "LDFLAGS" => "-L#{install_dir}/embedded/lib #{arch_flag}",
-            "CFLAGS" => "-I#{install_dir}/embedded/include #{arch_flag}",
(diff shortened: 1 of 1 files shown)


danielsdeleo:
@ksubrama did you need those for building ruby itself from source?

ksubrama:
I did but if these are problematic, I don't mind them being commented out.  I can define them in the ruby software project.

ksubrama:
@tyler-ball Can you paste the error you ran into - I'm surprised how extremely rare environment variables affect ChefDK given how we don't ever compile anything ever for windows so far.

tyler-ball:
@ksubrama and I talked in person - he agrees that this is the right thing to do.

danielsdeleo:
:+1: then

PR #600, opened 2016-01-11
add libmd.so.* to freebsd whitelist

lamont-granquist:
the current version of this library seems to be the one in the freebsd
source code.  there's old versions that lack sha-2 support at

http://martin.hinner.info/libmd/
https://github.com/epowers/libmd

without something better to pull into as a software defn, it seems
best to just use the base o/s.

i also looked for ways of pointing the software at openssl instead of
libmd, but couldn't find any.

### DIFF
--- a/lib/omnibus/health_check.rb
+++ b/lib/omnibus/health_check.rb
@@ -140,6 +140,7 @@ module Omnibus
       /libelf\.so/,
       /libkvm\.so/,
       /libprocstat\.so/,
+      /libmd\.so/
     ].freeze
 
     class << self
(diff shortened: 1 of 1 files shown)


lamont-granquist:
solaris and smartos also already whitelist this lib as well..

lamont-granquist:
@chef/omnibus-maintainers needs review

yzl:
:+1:

lamont-granquist:
tested in CI and it works, e.g.: http://manhattan.ci.chef.co/job/chef-12-build/75/architecture=i386,platform=freebsd-10,project=chef,role=builder/

PR #602, opened 2016-01-12
Be more paranoid when syncing files.

ksubrama:
If a file or a hardlink already exists at the destination,
we need to be careful and remote the existing file before
we perform the copy. A file at the destination can be
marked with r--r--r-- which will throw EACCESS if we blindly
copy from source to destination.

### DIFF
--- a/lib/omnibus/file_syncer.rb
+++ b/lib/omnibus/file_syncer.rb
@@ -106,9 +106,13 @@ module Omnibus
           # duplicate them, provided their source is in place already
           if hardlink? source_stat
             if existing = hardlink_sources[[source_stat.dev, source_stat.ino]]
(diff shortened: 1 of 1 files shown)


ksubrama:
There are still a few other use-cases (like source being a hard-link while destination is a read only directory or a symlink to a read only directory) that I'm unsure about that the code present fixes an issue we've encountered and I don't really feel like digging through the man pages of all our platforms for synthetic scenarios that we haven't encountered.

scotthain:
@curiositycasualty you ran into some of this oddness didn't you?

curiositycasualty:
Yeah but it looks like this should take care of it. Ideally, this would be a use-case added to harmony. @scotthain

schisamo:
:+1:

scotthain:
Sweet. :+1:

PR #603, opened 2016-01-14
Update submodules on git checkout

szechyjs:


ksubrama:
:+1:

scotthain:
I feel like this caused an issue at some point in the past? @chef/engineering-services

szechyjs:
Have we made any decisions on this yet?

I currently have to use the following hack in my software definitions.

```ruby
command 'git submodule update', env: env
```

ksubrama:
@schisamo @scotthain I am personally +1 on this and think it's essential.  git cache has received numerous bug-fixes recently and I don't believe this should cause us any breakage unless we have platforms that outright break on submodules (which is a different problem - they'll fail anyway).

rhass:
The `enable_submodules` property in Chef uses the same behavior. 
https://github.com/chef/chef/blob/master/lib/chef/provider/git.rb#L174
I am in favor of doing this to keep thing consistent. :+1: 

If problems do arise we can always make something like `update_submodules` as a separate parameter.

yzl:
:+1:

PR #607, opened 2016-01-19
Go back to using tar on windows

ksubrama:
Both 7z and bsdtar have issues on windows.

7z cannot extract and untar at the same time. You need to extract to a
temporary location and then extract again into project_dir.

7z also doesn't handle symlinks well. A symlink to a non-existent
location simply results in a text file with the target path written in
it. It does this without throwing any errors.

bsdtar will exit(1) if it is encounters symlinks on windows. So we can't
use shellout! directly.

bsdtar will also exit(1) and fail to overwrite files at the destination
during extraction if a file already exists at the destination and is
marked read-only. This used to be a problem when we weren't properly
cleaning an existing project_dir. It should be less of a problem now...
but who knows.

Add an option called :extract that allows us to run tar in any of these modes.
Also, by default, use "tar" on windows since it produces the most sensible
results in most scenarios.


jaym:
There's probably a really good reason to do this. Could you add that to the description of your PR or the commit message.

jaym:
Why do the symlinks matter? It seems having an empty file would be fine if the symlinks never get used. `lax_tar` seems like a really bad idea as you cannot be sure why something failed.

ksubrama:
Because it doesn't actually give you an empty file.  7zip produces a file with the symlink target printed in it.  Since the file is now present, other downstream users attempt to use it or compile it like a C file or run it and blow up.

danielsdeleo:
:+1:

lib/omnibus/software.rb
```diff
@@ -546,6 +560,18 @@ def with_standard_compiler_flags(env = {}, opts = {})
...
+      # optional flag.
+      if opts[:bfd_flags] && windows?
+        bfd_target = windows_arch_i386? ? "pe-i386" : "pe-x86-64"
+        compiler_flags.merge!(
+          {
+            "RCFLAGS" => "--target=#{bfd_target}",
+            "ARFLAGS" => "--target=#{bfd_target}",
+          }
+        )
+      end
```
schisamo: Can we please get some test coverage for this?

ksubrama: Done!

jaym:
the lax_tar stuff is a little concerning, but if there's no other way :+1:

PR #609, opened 2016-01-21
this autoprunes files with spaces

lamont-granquist:
these are most often spec fixtures in rubygems that have spaces in
the filenames.  we keep needing to add delete lines to remove
tests from gems before packaging.  this patch automates that.

### DIFF
--- a/lib/omnibus/packagers/solaris.rb
+++ b/lib/omnibus/packagers/solaris.rb
@@ -68,11 +68,11 @@ module Omnibus
       SCRIPT_MAP.each do |source, destination|
         source_path = File.join(project.package_scripts_path, source.to_s)
 
-        if File.file?(source_path)
-          destination_path = staging_dir_path(destination)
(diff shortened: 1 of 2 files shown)


lamont-granquist:
@chef/omnibus-maintainers assuming rubocop is happy...

schisamo:
:+1:

lamont-granquist:
so to capture discussion on slack...

this could omit a critical file that we need for runtime which has a space in it.

in that circumstance, though, we're hosed because we can't package files with spaces on solaris.

we should also catch that via some kind of test failure.

the alternative is to periodically break builds when we deploy gems with spec fixtures that have spaces in filenames, and then we have to remember why that happens and go fiddle around manually deleting the file.

so its creating busy work and failed builds, so that we know up front that we broke the build, when we should have tests that will capture that, and we're just completely f***ed anyway if the thing we're worried about ever really happens...

so i don't see the point in not doing this...

scotthain:
I'm cool with it - we should trust our tests :)

PR #611, opened 2016-01-22
Allow assets with non-md5 checksums to be cached in s3

ksubrama:
Also make omnibus cache missing detect missing software across all the various project definitions and configurations you have.


ksubrama:
@jaym @btm @chef/engineering-services

lib/omnibus/fetchers/net_fetcher.rb
```diff
@@ -130,6 +130,25 @@ def checksum
...
+    # If the user provides an md5 in the software definition, then that is
+    # used. Otherwise, the downloaded file is verified using the user
+    # specified digest and a fresh md5 checksum is calculated.
+    #
+    # Remember to call this function with s3 caching disabled if you are
+    # attempting to generate this md5 to look up this object in s3.
+    #
+    # @return [String]
+    #
+    def md5
```
jaym: This seems a little odd to me. A function like `md5` is very much a getter in brain. Having it go off and maybe download a file feels weird.

Related question:
What happens if your software definition only specifies a sha and then we try fetch that package from the cache? Are we going to download form the canonical source to get the hash and then from the s3 cache?

ksubrama: The object will still be named by its checksum (sha1).  This is just to populate the md5 field when uploading to s3.

Now that I think about it though - you're right.  It might be better to move this function into s3_cache and just have it calculate the md5.  digest_type was not being exposed and I was hoping to initially mirror the working of "checksum" with "actually get the md5" but that's more confusing that having the s3_cache module simply recalculate the md5 unconditionally when uploading to s3.

lamont-granquist: yeah +1 on it seeming odd

ksubrama:
@jaym I've simplified this now.

lib/omnibus/s3_cache.rb
```diff
@@ -75,9 +76,14 @@ def populate
           log.info(log_key) do
             "Caching '#{fetcher.downloaded_file}' to '#{Config.s3_bucket}/#{key}'"
           end
+          
+          # Fetcher has already verified the downloaded file in software.fetch.
+          # Compute the md5 from scratch because the fetcher may have been
+          # specified with a different hashing algorithm.
+          md5 = digest(fetcher.downloaded_file, :md5)
```
jaym: I think the checksum that was provided needs to be verified

jaym: nvm:
https://github.com/chef/omnibus/blob/master/lib/omnibus/fetchers/net_fetcher.rb#L86

jaym:
:+1:

PR #612, opened 2016-01-28
Optimize the crap out of windows

ksubrama:


lamont-granquist:
yeah enabling SSE2 is a 2001-era optimization.   :+1:

thommay:
Do we actually want to O3 rather than O2? Just thinking for supportability/stability's sake a slightly less agressive approach might be better

lamont-granquist:
ruby's default has been -O3 since 2009:

https://github.com/ruby/ruby/commit/62e7e47bdb6d5b4f13138bd3290ca3c40d3ddbab

thommay:
Huh. go ahead then! :+1:

lamont-granquist:
These would probably be good as generic options for everything other than AIX/Solaris?  If we're compiling for i586 without any SSE support that's a little bit awful...

schisamo:
@ksubrama remember to /cc @chef/omnibus-maintainers plz.

ksubrama:
Oh woah this got popular.  Sorry @schisamo - I should have marked this WIP.  I just wanted some builds going.

The optimization flags help in most places but some (like openssl) seem to dislike it and I'm currently debugging it.  I'll send this out once I'm confident that it won't break anything.

ksubrama:
Ok, I am ready to merge this.  OpenSSL was actually fine.  This is windows only for now - I'll make a different PR to see what happens if we enable SSE2 on other platforms next.

lamont-granquist:
While you're at it no-exec-stack flags on linux and all that jazz would be good, too, since I'm sure you have no other work to do...

PR #613, opened 2016-01-28
Do not modify CRLF when git caching

ksubrama:


lib/omnibus/git_cache.rb
```diff
@@ -129,16 +129,19 @@ def restore
...
       restore_me = false
-      cmd.stdout.each_line do |line|
-        restore_me = true if tag == line.chomp
+      begin
+        cmd = git_cmd(%Q(tag -l "#{tag}"))
+
+        cmd.stdout.each_line do |line|
+          restore_me = true if tag == line.chomp
+        end
+      rescue CommandFailed
```
jaym: why is this allowed to fail? Should we at least log something?

ksubrama: You're right.  I was being lazy about investigating why this call was using shellout instead of shellout! but it seems to be a historical artifact. I switched it to not catch any exceptions.

schisamo:
This needs to be tested HEAVILY before we merge. Every time the git cache logic is tweaked we have what feels like months of fallout.

jaym:
I'm okay with letting it bake before we merge. How about we float the chef branch onto this for a week or so?

That being said, I think this is the right thing to do.

stevendanna:
This looks good to me but definitely needs a test.

ksubrama:
I have sent a few chef-fips builds through this across cache invalidations and it seems to not have incurred any fall out.  A formal test for this would involve a functional test and verifying line endings.  It only affects windows per-se because I don't believe we have any other platform where we need to perform new-line conversion (we don't have any windows centric code with \r\n in them).

Informally, I actually caught this on chef-fips pipeline when build generated a fips_premain.c file (or something like that) and calculated the SHA256 for it.  I believe it only injected \ns when generating it.  The git cache mangled it when saving/restoring it and now it had \r\n and the FIPS validating compiler failed to build ruby from the "tainted" source.

I'll send out a PR to float both chef and chef-fips on this for a week.

danielsdeleo:
:+1: from me as long as we do enough test builds to be confident. Could also just merge and revert if it does anything unexpected.

ksubrama:
Well...  *I'm* confident that the worst that could happen is we simply need to revert and expire caches if this change turns out to be bad on some weird platform with some old git that doesn't support this or something.  Whether you should trust me is a different matter.

danielsdeleo:
If we have the capability to completely nuke anything this touches from the builders in the case we needed to revert, then I'd just go for it.

ksubrama:
Leeeeroy!

PR #614, opened 2016-01-28
Print NetFetcher retries at the info level

btm:
It would be really helpful to know if we had retried a download while
debugging intermittent failures.


jaym:
:+1:

danielsdeleo:
:+1:

schisamo:
I'm not against this change but we should be seeing debug level output in our CI system. We execute all builds with `-l internal` which includes `debug` level output:
https://github.com/chef-cookbooks/opscode-ci/blob/master/libraries/omnibus_build_pipeline.rb#L296-L303

btm:
added the exception to the output and force pushed.

PR #615, opened 2016-01-28
kill ourselves and all our children

lamont-granquist:
Extend the SIGINT handler so that we nuke our own process group from orbit.

This may fix some issues with cancelling jobs in jenkins and not having the process tree cleaned up and subsequent jobs failing.

Does not address windows.  Similar code from mixlib-shellout could probably be stolen to fix windows...

### DIFF
--- a/bin/omnibus
+++ b/bin/omnibus
@@ -2,7 +2,7 @@
 
 # Trap interrupts to quit cleanly. See
 # https://twitter.com/mitchellh/status/283014103189053442
-Signal.trap('INT') { exit 1 }
+Signal.trap('INT') { Process.kill(:KILL, -Process.getpgid(Process.pid()) }
 
 $:.push File.expand_path('../../lib', __FILE__)
 $stdout.sync = true
--- a/bin/omnibus
+++ b/bin/omnibus
@@ -1,8 +1,20 @@
 #!/usr/bin/env ruby
 
-# Trap interrupts to quit cleanly. See
-# https://twitter.com/mitchellh/status/283014103189053442
-Signal.trap('INT') { Process.kill(:KILL, -Process.getpgid(Process.pid()); exit 1 }
+# become our own process leader so we can kill our process tree
+unless Gem.win_platform?
+  Process.setpgid(0,0)
+end
+
+Signal.trap('INT') do
+  unless Gem.win_platform?
+    Process.kill(:INT, -Process.getpgid(Process.pid()))
+    sleep 1
+    Process.kill(:KILL, -Process.getpgid(Process.pid()))
+  end
+  # Trap interrupts to quit cleanly. See
+  # https://twitter.com/mitchellh/status/283014103189053442
+  exit 1
+end
 
 $:.push File.expand_path('../../lib', __FILE__)
 $stdout.sync = true
--- a/bin/omnibus
+++ b/bin/omnibus
@@ -2,7 +2,7 @@
 
 # Trap interrupts to quit cleanly. See
 # https://twitter.com/mitchellh/status/283014103189053442
-Signal.trap('INT') { Process.kill(:KILL, -Process.getpgid(Process.pid()) }
+Signal.trap('INT') { Process.kill(:KILL, -Process.getpgid(Process.pid()); exit 1 }
 
 $:.push File.expand_path('../../lib', __FILE__)
 $stdout.sync = true


lamont-granquist:
@chef/omnibus-maintainers this might help with jenkins and stopping jobs.

jaym:
:+1: I feel like @mwrock might have ideas for windows. He recently fixed a bug of mine somewhere related to not killing children

mwrock:
its probably the same stuff @lamont-granquist referred to in mixlib-shellout that had my fix to kill all child processes

schisamo:
Is it possible to add some additional Aruba test coverage?

lamont-granquist:
its pretty difficult to add coverage for this.  i can write some entirely meaningless unit tests.  functional tests would also be difficult to fork off a real omnibus process to fork off a real badly-behaved child process that SIGINT of the parent would cause the child to not clean up, and then be able to have the grandparent process observe the failure in order to go red when this change was backed out...

lamont-granquist:
plus i have no idea what aruba tests are...

danielsdeleo:
Yeah, this stuff is hard to test. For shellout, the tests get pretty nasty, having to run things in loops to ensure we create certain race conditions and so on: https://github.com/chef/mixlib-shellout/blob/master/spec/mixlib/shellout_spec.rb#L1016-L1281

lamont-granquist:
Yeah that's what I'd start looking at.  But then aruba.  And making omnibus do this through really executing it and having a software definition that would function like those test in shell_out.  And that returns -ENOTIMEFORTHAT for me...  Plus we're not talking about commands timing out and just making sure they exit, but some kind of theoretically stuck processes on jenkins that a SIGINT to the omnibus proc isn't cleaning up, so the software defn would have to mangle its own SIGINT handler and then the test process would need to be able to find it (portably so the tests can run on other systems) and see it still running and kill it...

lamont-granquist:
So a bit more importantly, I tried adding `signal('SIGINT', 'IGNORE')` to the start of the handler but that failed (can't recall the error offhand right now).  Supposedly ruby interrupt handlers need to be re-entrant, but I clearly do **not** observe that the SIGINT to the PGRP is re-triggering the signal handler (and winding up blowing up the stack in sigints) but that it proceeds and dies when it SIGKILLs itself.  I find that a bit odd, but it works this way.

schisamo:
@lamont-granquist Aruba is what we use to test Omnibu's CLI interface. The existing test coverage uses the Cucumber dialect but it also supports RSpec and Minitest:
https://github.com/chef/omnibus/tree/master/features

We also use Aruba to test the CLI interfaces for __test-kitchen__ and __delivery-cli__:
https://github.com/test-kitchen/test-kitchen/tree/master/features
https://github.com/chef/delivery-cli/tree/master/features

Since we are modifying the behavior of the CLI interface it seems worthwhile to add test coverage.

lamont-granquist:
Its not CLI though its process tree management.

rhass:
Sign the DCO and merge this? This seems like it would solve all kinds of issues on Windows.

stevendanna:
> Can we sign the DCO and merge this? This seems like it would solve all kinds of issues on Windows.

I'm in favor of merging this, but how would it help on Windows? All of the new code is skipped on Windows.

mwrock:
This would not help Windows. Windows implements process groups (and signaling) differently. On windows you can start a process to be the root of a new process group by passing a `CREATE_NEW_PROCESS_GROUP` flag to the `CreateProcessW` call in mixlib-install. However there is no easy equivilent of sending s `SIGTERM` to a windows process group. One can call `GenerateConsoleCtrlEvent` to send a `ctrl+BREAK` to a process group which is the moral equivilent of a `SIGINT` but if that is not honored by the process, one has to recursively terminate all child processes of the offending process to guarantee no orphan processes are left behind.

I recently [implemented this in rust](https://github.com/habitat-sh/habitat/blob/9e6f4e23103baae6de1db2a75fbbc4abd48bd856/components/core/src/os/process/windows.rs#L184) for the Habitat supervisor so its fresh in my mind.

lamont-granquist:
@mwrock just implement whatever the easiest equivalent is of sending sigkill to the entire process tree.  the use of sigint here is unnecessary, its included only by convention because its easy.

curiositycasualty:
👍  for the title alone

PR #617, opened 2016-02-03
Fix bff cleanup

scotthain:
@chef/engineering-services 
@chef/omnibus-maintainers 

Long story short - the due to the way `mkinstallp` works, the proto directory needs to be owned by root. Since we weren't chowning this directory back to the effective build user, it was not being cleaned up by the cleanup method in `base.rb`. This workaround sets it back to the build user so it can be cleaned up appropriately.

### DIFF
--- a/lib/omnibus/packagers/bff.rb
+++ b/lib/omnibus/packagers/bff.rb
@@ -208,7 +208,10 @@ module Omnibus
       # This implies that if we are in /tmp/staging/project/dir/things,
       # we will chown from 'project' on, rather than 'project/dir', which leaves
       # project owned by the build user (which is incorrect)
-      shellout!("sudo chown -R 0:0 #{File.join(staging_dir, project.install_dir.match(/^\/?(\w+)/).to_s)}")
+      # First - let's find out who we are.
+      original_uid = shellout!("id -u").stdout.chomp
+      original_gid = shellout!("id -g").stdout.chomp
+      shellout!("sudo chown -Rh 0:0 #{File.join(staging_dir, project.install_dir.match(/^\/?(\w+)/).to_s)}")
       log.info(log_key) { "Creating .bff file" }
 
       # Since we want the owner to be root, we need to sudo the mkinstallp
@@ -229,6 +232,9 @@ module Omnibus
       FileSyncer.glob(File.join(staging_dir, 'tmp/*.bff')).each do |bff|
         copy_file(bff, File.join(Config.package_dir, create_bff_file_name))
       end
+
+      # chown back to original user's uid/gid so cleanup works correctly
+      shellout!("sudo chown -Rh #{original_uid}:#{original_gid} #{staging_dir}")
     end
 
     #
--- a/lib/omnibus/packagers/bff.rb
+++ b/lib/omnibus/packagers/bff.rb
@@ -209,8 +209,6 @@ module Omnibus
       # we will chown from 'project' on, rather than 'project/dir', which leaves
       # project owned by the build user (which is incorrect)
       # First - let's find out who we are.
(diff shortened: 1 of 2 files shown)


lib/omnibus/packagers/bff.rb
```diff
@@ -229,6 +232,9 @@ def create_bff_file
       FileSyncer.glob(File.join(staging_dir, 'tmp/*.bff')).each do |bff|
         copy_file(bff, File.join(Config.package_dir, create_bff_file_name))
       end
+
+      # chown back to original user's uid/gid so cleanup works correctly
+      shellout!("sudo chown -Rh #{original_uid}:#{original_gid} #{staging_dir}")
```
schisamo: I'm wondering if this should be in an ensure block so we actually ensure ownership is changed back...even in the event  `mkinstallp` causes an exception to bubble up. 

```ruby
def create_bff_file
  ... # do original stuff here
ensure
  # chown back to original user's uid/gid so cleanup works correctly
  original_uid = shellout!("id -u").stdout.chomp
  original_gid = shellout!("id -g").stdout.chomp
  shellout!("sudo chown -Rh #{original_uid}:#{original_gid} #{staging_dir}")
end
```

scotthain: I'm not sure it totally matters, if there is an exception we want to keep the directory intact in the state it was in when it failed, no?

sersut: I think if something fails we still want to chown it back for cleanup, especially since this is all in a tmpdir and we will need to do a new build anyways :)

schisamo:
:+1:

lib/omnibus/packagers/bff.rb
scotthain: My only reason against that would be leaving the builder state exactly how it was when it failed, to help debug... I'm willing to be overruled if you and @schisamo feel strongly we should chown back :)

schisamo: So I've been bothered by our use of `sudo` in this packager from the start. I wonder if we should re-examine our use of `mkinstallp` and just have the packager perform the steps as mentioned here:
http://compgroups.net/comp.unix.aix/why-does-mkinstallp-insist-on-being-run-as-root/1250985

If not I think we should ensure we always chown things back.

scotthain: @schisamo Unfortunately, doing the manual steps cause the files inside the resulting package to be owned by the existing user when they are installed on a target machine. The chown/sudo 'fix' was actually set up to fix a bug where packages were created with my username and wouldn't install correctly on target machines. Short of patching mkinstallp (which I seriously don't think we should do) I don't see a solution other than sudo.

yzl: I too vote for ensuring that we put everything back the way it was.  (I'd actually vote for a default of "put it back the way it was" and a debugging option for "leave everything in the state it was in when the build failed", but that's beyond scope.)

lib/omnibus/packagers/bff.rb
```diff
@@ -208,7 +208,10 @@ def create_bff_file
       # This implies that if we are in /tmp/staging/project/dir/things,
       # we will chown from 'project' on, rather than 'project/dir', which leaves
       # project owned by the build user (which is incorrect)
-      shellout!("sudo chown -R 0:0 #{File.join(staging_dir, project.install_dir.match(/^\/?(\w+)/).to_s)}")
+      # First - let's find out who we are.
+      original_uid = shellout!("id -u").stdout.chomp
+      original_gid = shellout!("id -g").stdout.chomp
```
schisamo: Since these `original_*` vars are not used until the ensure block it's best to declare them down there.

yzl:
:+1:

PR #619, opened 2016-02-04
force builds nil versions

lamont-granquist:
setting to 0.0.0 in the cache is always going to get the caching
wrong.  this force builds every time which at least gets the
caching correct.

### DIFF
--- a/lib/omnibus/software.rb
+++ b/lib/omnibus/software.rb
@@ -1,5 +1,5 @@
 #
-# Copyright 2012-2017, Chef Software Inc.
+# Copyright 2012-2018, Chef Software Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


lamont-granquist:
existing behavior is clearly going to bug the cache.  one possible fix here.  there appear to be no tests around this behavior and i hit my timebox chasing this squirrel.

lib/omnibus/software.rb
```diff
@@ -906,7 +906,12 @@ def source_type
     #
     def build_me
       if Config.use_git_caching
-        if project.dirty?
+        if !resolved_version
+          log.info(log_key) do
+            "Forcing a build because resolved version is nil"
+          end
+          execute_build
```
stevendanna: Do we want to dirty the cache in this case?

lib/omnibus/software.rb
```diff
@@ -833,17 +833,17 @@ def version_guid
       fetcher.version_guid
     end
 
+    def resolved_version
```
stevendanna: A little hesitant to call this resolved version because there is a resolved_version in the fetcher as well.

lib/omnibus/software.rb
lamont-granquist: The doc comment for execute_build indicates that it does?

stevendanna: I'll need to take a new look at that code.  I believe that execute_build will certainly build without the git cache and make a new project in the cache, but I don't see anywhere that would call project.dirty! to ensure that the following softwares also build without the cache.

lamont-granquist:
@stevendanna fixed the stuff you mentioned here.

tas50:
@lamont-granquist is this something we want to keep open?

lamont-granquist:
AFAIK its good to merge.

It might be the fix to the issues with the git caching when switching between major versions of chef-client.  It certainly is a big hole in how git caching works.

lamont-granquist:
YOLO

PR #623, opened 2016-02-08
There is no restriction on dashes in the base package name or identifier

scotthain:
@chef/engineering-services @chef/omnibus-maintainers 

I ran through a full package test locally on my Mac successfully, and it looks like this is correct based on Apple's documentation: https://developer.apple.com/library/mac/documentation/FileManagement/Conceptual/understanding_utis/understand_utis_conc/understand_utis_conc.html

### DIFF
--- a/lib/omnibus/packagers/pkg.rb
+++ b/lib/omnibus/packagers/pkg.rb
@@ -262,14 +262,14 @@ module Omnibus
     # @return [String]
     #
     def safe_base_package_name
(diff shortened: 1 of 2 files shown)


yzl:
:+1: 

(It would be nice to try out a package with a dash in its name on the oldest version of OS X we use, but I suspect it will be ok.)

schisamo:
:+1:

PR #626, opened 2016-02-10
Fallback to system bash for windows

jaym:
We don't want to rely on the build tools being part of the omnibus package.


schisamo:
:+1: 

/cc @chef/omnibus-maintainers

yzl:
:+1:

PR #628, opened 2016-02-17
Disable SSE optimization flags - OpenSSL dislikes it

ksubrama:


lib/omnibus/software.rb
```diff
@@ -571,11 +571,17 @@ def with_standard_compiler_flags(env = {}, opts = {})
           freebsd_flags
         when "windows"
           arch_flag = windows_arch_i386? ? "-m32" : "-m64"
-          opt_flag = windows_arch_i386? ? "-march=i686" : "-march=x86-64"
+          #opt_flag = windows_arch_i386? ? "-march=i686" : "-march=x86-64"
```
jaym: I think having your comments is fine, but you should just delete the commented out code. You can always come back and get it

lib/omnibus/software.rb
```diff
@@ -571,11 +571,17 @@ def with_standard_compiler_flags(env = {}, opts = {})
...
           {
             "LDFLAGS" => "-L#{install_dir}/embedded/lib #{arch_flag}",
-            # If we're happy with these flags, enable SSE for other platforms running x86 too.
-            "CFLAGS" => "-I#{install_dir}/embedded/include #{arch_flag} -O3 -mfpmath=sse -msse2 #{opt_flag}"
+            # If we're happy with these flags, enable SSE for other platforms
```
jaym: Is this still true?

jaym:
I think this puts us closer to what all the knapbuild things did, so I'm :+1:

lib/omnibus/software.rb
btm: is there a comment about why the -march is removed too?

lib/omnibus/software.rb
ksubrama: Well - that's there for Lamont :)  He wanted to do it on solaris and other slow platforms.

lib/omnibus/software.rb
ksubrama: It was overriding openssl's explicit -march i486 flag.  I'll add a comment to that effect too.

@jaym I was leaving the code in for context as to what cannot be done without investigation.  It's easy for someone to accidentally enable a flag without checking history to see if that flag was disabled.

lamont-granquist:
can we enable SSE everywhere but disable it on openssl?

danielsdeleo:
Yeah, seems likely that OpenSSL is a special case.

ksubrama:
Withdrawn given the above PR in omnibus-software.

PR #632, opened 2016-03-03
Add update_config_guess DSL method

sersut:
This PR adds a new dsl method to easily update a software's config_guess from a config_guess installed with a software dependency.

Replaces https://github.com/chef/omnibus/pull/592.
Implements the style discussed in https://github.com/chef/omnibus/pull/592#discussion_r53094053. 

This will be followed up by a PR in omnibus-software that adds a config_guess software definition.

/cc: @scotthain, @chef/omnibus-maintainers

### DIFF
--- a/lib/omnibus/builder.rb
+++ b/lib/omnibus/builder.rb
@@ -644,6 +644,35 @@ module Omnibus
     expose :sync
 
     #
+    # Helper method to update config_guess in the software's source
+    # directory.
+    # You should add a dependency on the +config_guess+ software definition if you
+    # want to use this command.
+    # @param [Hash] options
+    #   Supported options are:
+    #     target [String] subdirectory under the software source to copy
+    #       config.guess.to. Default: "."
+    #     install [Array<Symbol>] parts of config.guess to copy.
+    #       Default: [:config_guess, :config_sub]
+    def update_config_guess(target: ".", install: [:config_guess, :config_sub])
+      build_commands << BuildCommand.new("update_config_guess `target: #{target} install: #{install.inspect}'") do
+        config_guess_dir = "#{install_dir}/embedded/lib/config_guess"
+        %w{config.guess config.sub}.each do |c|
+          unless File.exist?(File.join(config_guess_dir, c))
+            raise "Can not find #{c}. Make sure you add a dependency on 'config_guess' in your software definition"
+          end
+        end
+
+        destination = File.join(software.project_dir, target)
+        FileUtils.mkdir_p(destination)
+
+        FileUtils.cp_r("#{config_guess_dir}/config.guess", destination) if install.include? :config_guess
+        FileUtils.cp_r("#{config_guess_dir}/config.sub", destination) if install.include? :config_sub
+      end
+    end
+    expose :update_config_guess
+
+    #
     # @!endgroup
     # --------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


lib/omnibus/builder.rb
```diff
@@ -644,6 +644,39 @@ def sync(source, destination, options = {})
...
     #
+    # Helper method to update config_guess in the software's source
+    # directory.
+    # @param [Hash] options
+    #   Supported options are:
+    #     target [String] subdirectory under the software source to copy
+    #       config.guess.to. Default: "."
+    #     install [Array<Symbol>] parts of config.guess to copy.
+    #       Default: [:config_guess, :config_sub]
+    def update_config_guess(target: ".", install: [:config_guess, :config_sub])
```
scotthain: This should default to updating both unless you specify, per @lamont-granquist's comment.

scotthain:
Ping @edolnx

lib/omnibus/builder.rb
sersut: I think it defaults to updating both. I just didn't want to add an `:all` option :smile:

lib/omnibus/builder.rb
```diff
@@ -644,6 +644,39 @@ def sync(source, destination, options = {})
     expose :sync
 
     #
+    # Helper method to update config_guess in the software's source
+    # directory.
+    # @param [Hash] options
```
schisamo: We should update the comment to mention a `config_guess` dependency is required in any software def that depends on this behavior. See the `bundle` method for an example:
https://github.com/chef/omnibus/blob/94b293b711e3372e04c22f92974d2248a9e60330/lib/omnibus/builder.rb#L317-L318

lib/omnibus/builder.rb
```diff
@@ -644,6 +644,39 @@ def sync(source, destination, options = {})
...
+    #       config.guess.to. Default: "."
+    #     install [Array<Symbol>] parts of config.guess to copy.
+    #       Default: [:config_guess, :config_sub]
+    def update_config_guess(target: ".", install: [:config_guess, :config_sub])
+      build_commands << BuildCommand.new("update_config_guess `target: #{target} install: #{install.inspect}'") do
+        destination = File.join(software.project_dir, target)
+        FileUtils.mkdir_p(destination)
+
+        config_guess_dir = FileSyncer.glob("#{Config.source_dir}/*/").select do |d|
+          d =~ /config_guess/
```
stevendanna: Why are we doing this via a regex rather than as part of the glob?  Also, what do we expect these to be named.  The error messages below seem to indicate that the directory will be named `config_guess-SOMETHING` but this would pick up much more than that.

lib/omnibus/builder.rb
sersut: Ah. Nice catch! Will do :+1:

lib/omnibus/builder.rb
sersut: Just realized this yesterday @stevendanna. Looks like I forgot how `cache/src` layout :smile: will fix soon.

sersut:
@stevendanna updated and also added an example of how to use this https://github.com/chef/omnibus-software/pull/614. LMK what you think :smile:

schisamo:
:+1:

yzl:
:+1:

PR #633, opened 2016-03-09
Update version and CHANGELOG.md for 5.0.2 release.

sersut:
This PR prepares us for 5.0.2 release. 

We are skipping 5.0.1 because:
- We will be adding omnibus as a dependency of omnibus-software because omnibus-software is going to start using `update_config_guess` method.
- We want to ensure that we depend on a version that has `update_config_guess` in it. 
- Since many projects pull omnibus from github we can end up in a scenario where you are getting a version without `update_config_guess` method if we release 5.0.1.

/cc: @chef/omnibus-maintainers


danielsdeleo:
:+1:

CHANGELOG.md
```diff
@@ -1,6 +1,32 @@
...
+- Use -O2 when building with standard compiler flags. (#591)
+- Cache software sources under `.../src/<software name>/<package>`. (#597)
+- Add libmd.so.* to freebsd whitelist. (#600)
+- Remove existing links in the destination when syncing files. (#602)
+- Skip adding DEBIAN directory to md5sums file. (#595)
+- Autoprunes files with spaces on Solaris. (#609)
+- Allow assets with non-md5 checksums to be cached in s3. (#611)
+- Print NetFetcher retries at the info level. (#614)
+- Do not modify CRLF when git caching. (#616)
+- Ensurewe always swap chown back to default. (#617)
```
tas50: Missing a space here

lib/omnibus/version.rb
```diff
@@ -15,5 +15,5 @@
 #
 
 module Omnibus
-  VERSION = '5.0.1'
+  VERSION = '5.0.2'
```
schisamo: I think this should bump to `5.1.0` since we are adding new, backward compatible, features (namely the DSL methods).

sersut: :+1: Will do @schisamo

PR #635, opened 2016-03-11
License reporting in omnibus

sersut:
This PR adds omnibus the ability to collect and report license information for projects and software components that are being included. I will try my best to describe this feature below but if you have more detailed questions, hit me directly and we can chat. 

This features targets two main scenarios:

* As an enterprise user; I would like to see 1) "the list of software components", 2) "the licenses these components are using" and 3) "the actual files that are referred by these licenses" for the package that I am installing, so that I can check these against our companies open source licensing policy and make sure we are compliant.
* As a developer, I would like to be informed about the licenses of the software components that are going into the package that I am building, so that I can ensure we are not including a software component with a license that is incompatible with our license.

This PR contains the MVP of this feature. You can see [this gist](https://gist.github.com/sersut/e1012c38e55a3890232b) with an example output created with https://github.com/chef/omnibus-software/pull/620 and https://github.com/chef/chef-server/pull/775.

Here is a high level summary of this feature:

- Adds 2 new DSL methods to software:
  - `license`: The name of the license software is using.
  - `license_file`: File that are part of the license for the software. Can be a partial path relative to the software's source directory or a URL. Can be used multiple times if license has more than 1 file.
- Adds 3 new DSL methods to project:
  - `license`: The name of the license project is using.
  - `license_file`: File that contains the text of the license project is using.
  - `license_file_path`: Path of the file that will be created under `install_dir` which will contain license information for the project.
- Adds a new post-build stage to the build command. During this stage omnibus does two things:
  - It copies the license files specified by software definitions to `#{install_dir}/LICENSES`
  - It creates a file at `license_file_path` specified by the project that contains:
    - project's name, version and license name. 
    - license name and list of file paths that belong to the license for each software
  - If you use the special license type `:project_license` in software definitions, we skip adding that software's license information to the main license file. We use this for things like, `preparation`, `private-chef-cookbooks`, `private-chef-ctl`, etc..
- Updates the version manifest files with the license name of each software.
There are many features that are missing to fulfill the above mentioned scenarios. Some that we have thought of are:
* Add warnings to omnibus build output for important cases like software config that is missing license, license files that do not exist anymore, etc. **(next item)**
* Add logic to catch licenses of hidden dependencies like gems that are being installed from rubygems during `bundle install` command runs.

Comments, questions, suggestions, error cases I have missed are much appreciated :smile: 

/cc: @chef/omnibus-maintainers, @jamesc

### DIFF
--- a/lib/omnibus/software.rb
+++ b/lib/omnibus/software.rb
@@ -328,6 +328,46 @@ module Omnibus
     expose :default_version
 
     #
+    # Set or retrieve the {#license} of the software to build.
+    #
+    # @example
+    #   license 'Apache 2.0'
+    #
+    # @param [String] val
+    #   the license to set for the software.
+    #
+    # @return [String]
+    #
+    def license(val = NULL)
+      if null?(val)
+        @license || 'Unspecified'
+      else
+        @license = val
+      end
+    end
+    expose :license
+
+    #
+    # Set or retrieve the location of a {#license_file}
+    # of the software.  It can either be a relative path inside
+    # the source package or a URL.
+    #
+    #
+    # @example
+    #   license_file 'LICENSES/artistic.txt'
+    #
+    # @param [String] val
+    #   the location of the license file for the software.
+    #
+    # @return [String]
+    #
+    def license_file(file)
+      license_files << file
+      license_files.dup
+    end
+    expose :license_file
+
+    #
     # Evaluate a block only if the version matches.
     #
     # @example
@@ -727,7 +767,9 @@ module Omnibus
                                    source_type: source_type,
                                    described_version: version,
                                    locked_version: Fetcher.resolve_version(version, source),
-                                   locked_source: source})
+                                   locked_source: source,
+                                   license: license
+      })
     end
 
     #
@@ -768,6 +810,17 @@ module Omnibus
     #
     def whitelist_files
       @whitelist_files ||= []
+    end
+
+    #
+    # The list of license files for this software
+    #
+    # @see #license_file
+    #
+    # @return [Array<String>]
+    #
+    def license_files
+      @license_files ||= []
     end
 
     #
--- a/lib/omnibus.rb
+++ b/lib/omnibus.rb
@@ -62,6 +62,7 @@ module Omnibus
   autoload :Templating,       'omnibus/templating'
   autoload :ThreadPool,       'omnibus/thread_pool'
   autoload :Util,             'omnibus/util'
+  autoload :Licensing,        'omnibus/licensing'
 
   autoload :GitFetcher,  'omnibus/fetchers/git_fetcher'
   autoload :NetFetcher,  'omnibus/fetchers/net_fetcher'
--- a/lib/omnibus/licensing.rb
+++ b/lib/omnibus/licensing.rb
@@ -1,0 +1,227 @@
+#
+# Copyright 2015 Chef Software, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'uri'
+require 'fileutils'
+
+module Omnibus
+  class Licensing
+    include Logging
+
+    OUTPUT_DIRECTORY = "LICENSES".freeze
+
+    class << self
+      # @see (Licensing#create!)
+      def create!(project)
+        new(project).create!
+      end
+    end
+
+    #
+    # The project to create licenses for.
+    #
+    # @return [Project]
+    #
+    attr_reader :project
+
+    #
+    # @param [Project] project
+    #   the project to create licenses for.
+    #
+    def initialize(project)
+      @project = project
+    end
+
+    #
+    # Creates the license files for given project.
+    # It is assumed that the project has already been built.
+    #
+    # @return [void]
+    #
+    def create!
+      prepare
+      create_software_license_files
+      create_project_license_file
+    end
+
+    #
+    # Creates the required directories for licenses.
+    #
+    # @return [void]
+    #
+    def prepare
+      FileUtils.rm_rf(output_dir)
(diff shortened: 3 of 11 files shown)


lib/omnibus/licenses.rb
```diff
@@ -0,0 +1,206 @@
...
+    #
+    def create_software_license_files
+      license_map.each do |name, values|
+        license_files = values[:license_files]
+
+        license_files.each do |license_file|
+          if license_file && is_local(license_file)
+            input_file = File.expand_path(license_file, values[:project_dir])
+            output_file = license_package_location(name, license_file)
+            FileUtils.cp(input_file, output_file)
```
stevendanna: In your PR description you mention:

> Add warnings to omnibus build output for important cases like software config that is missing license, license files that do not exist anymore, etc. (next item)

Does this currently fail with an error if the default license file doesn't actually exist inside the project?  If so, I think we'll want the warning/error handing for the case of a missing license file in this feature before merge so we don't break existing projects too badly.

stevendanna:
This is a necessary and important feature.  Thanks for working on it.

I'm wondering if license and license_file should actually be part of the "version" rather than the software directly to better handle cases where the license changes between versions.

A nice-to-have that I don't think need to be in the first cut of this would be a project-level exclusion list that would fail the build with an error if any softwares used licenses on the list.  This would above

lib/omnibus/licenses.rb
sersut: This should not have an affect on existing projects. If a project does not have any licensing information we will create a `LICENSE` file under its install directory that say `"This project has an 'Unspecified' license."` But we can turn it off if that is undesirable behavior. 

The only case that will fail is when the specified `license_file` is not there and I will handle that case now :smile:.

sersut:
Handling changing licenses between versions is a real case and we should handle it. However I do not think that happens often and I do not want to force us to add same license information to many versions.

I will suggest we add `license` and `license_file` to `version` as well and it will override the licensing information specified at the software level if used. How does that sound @stevendanna ?

 :+1: about the exclusion list feature.

stevendanna:
> I will suggest we add license and license_file to version as well and it will override the licensing information specified at the software level if used. How does that sound @stevendanna ?

Yes, something like that sounds perfect.  Much like how the `source` attributes work.

sersut:
:+1:

sersut:
@stevendanna turns out that you can pretty much use any other DSL method in a way the `source` works in the block you are giving to `version`. See https://github.com/chef/omnibus/blob/master/lib/omnibus/software.rb#L353-L355

I have tried this for `ruby.rb` and it worked nicely:

```

File: lib/omnibus/licenses.rb [hunk 2]
```diff
@@ -0,0 +1,206 @@
...
+    # Output directory to create the licenses in.
+    #
+    def output_dir
+      File.expand_path(OUTPUT_DIRECTORY, project.install_dir)
+    end
+
+    #
+    # Returns if the given path to a license is local or a remote url.
+    #
+    def is_local(license)
```
schisamo: This should be `local?`.

lib/omnibus/licenses.rb
```diff
@@ -0,0 +1,206 @@
...
+    #
+    # Summary of the licenses included by the softwares of the project.
+    # It is in the form of:
+    # ...
+    # This product bundles python 2.7.9,
+    # which is available under a "Python" License.
+    # For details, see:
+    # /opt/opscode/LICENSES/python-LICENSE
+    # ...
+    #
```
schisamo: Let's make all the comments/documentation follow the [YARD](http://yardoc.org/) format like the rest of the code base.

lib/omnibus/licenses.rb
```diff
@@ -0,0 +1,206 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'uri'
+require 'fileutils'
+
+module Omnibus
+  class Licenses
```
schisamo: The plurality of this class feels weird to me. Would `License` or `Licensing` make more sense?

lib/omnibus/software.rb
```diff
@@ -328,6 +328,46 @@ def default_version(val = NULL)
...
+    # @param [String] val
+    #   the location of the license file for the software.
+    #
+    # @return [String]
+    #
+    def license_file(file)
+      license_files << file
+      license_files.dup
+    end
+    expose :license_file
```
schisamo: Would it be more appropriate for the DSL to expose a block based function like `version`:

```ruby
license 'Apache 2.0' do
  file 'LICENSE'
end

license 'BSD' do
  file 'BSDL'
  file 'COPYING'
  file 'LEGAL'
end
```

This would obviously require a `Omnibus::License` class to instance_eval against but it could easily be used with both `Project` and `Software`.

schisamo:
It would eventually be nice to totally fail a build if a dependency (or more likely a transitive dep) uses an undesirable license.

lib/omnibus/licenses.rb
sersut: I :heart: `Licensing`.

lib/omnibus/licenses.rb
sersut: :+1:

lib/omnibus/licenses.rb
sersut: :+1:

lib/omnibus/software.rb
sersut: I am not sure what it gets us other than (maybe) slightly better syntax. It would be helpful if we needed to define two licenses for a software or project and separate their files.

For version we have a great need since we would like to differentiate other things based on version. But here I am more in favor of flat DSL methods.

adamedx:
Great to see this @sersut. It looks good, I'd like to know if we can add the following in a later commit: `license_public_location`. This would be an Interent-accessible uri to the license in the snapshot of the source code we're shipping. It's possible this would be a pain to maintain, in which case we could simply link to the public source code repo which would be far more stable. 

We've had users ask for that public location so they can assess the license for themselves at the source, which is why I'm asking.

I don't think this blocks mvp at all, but I'd like to know if it's in the plan. I'll keep looking at the rest.

sersut:
Nice catch @scotthain will do. 

@adamedx uri's are currently supported. No need to have a different DSL method for it.

schisamo:
Let's squash this feature branch down into a single commit before merge. It would nice to have the commit message be nice and descriptive, potentially using a subset of your awesome PR description.

Other than that this looks great @sersut :+1:

stevendanna:
:+1:

PR #637, opened 2016-03-16
Download remote license files into the LICENSES directory

sersut:
The goal of the license reporting feature in omnibus is to include all the licenses being used by the project in `LICENSES` directory. This PR makes it so that we fetch remote licenses. 

I have reused some of the code from net_fetcher to do this. 

I have also introduced a `licensing_warning` method which I am planning to change later to fail the build when a CLI option called `fail_the_build_when_there_is_licensing_issues` is set :smile: 

@chef/omnibus-maintainers

### DIFF
--- a/lib/omnibus/download_helpers.rb
+++ b/lib/omnibus/download_helpers.rb
@@ -1,0 +1,137 @@
+#
+# Copyright 2015 Chef Software, Inc.
+#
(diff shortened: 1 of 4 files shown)


jamesc:
+1 Looks good

scotthain:
:+1: awesome!

PR #638, opened 2016-03-16
Licensing warnings

sersut:
This PR adds some warnings to the build log so that we can see when licensing information is having issues. We will be introducing the option to convert these warnings to failures in a later PR. 

Added checks are:
* Warn when project / software licensing information does not exist.
* Warn when project / software doesn't point to any license files.
* Warn when project / software license files are missing.
* Warn when license set is not a standard license.

- [x] Incoming tests.

/cc: @chef/omnibus-maintainers, @jamesc

### DIFF
--- a/lib/omnibus/licensing.rb
+++ b/lib/omnibus/licensing.rb
@@ -55,6 +55,7 @@ module Omnibus
     #
     def create!
       prepare
+      validate_license_info
       create_software_license_files
       create_project_license_file
     end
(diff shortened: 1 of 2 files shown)


lib/omnibus/licensing.rb
```diff
@@ -70,6 +71,49 @@ def prepare
...
     #
+    # Inspects the licensing information for the project and the included
+    # software components. Logs the found issues to the log as warning.
+    #
+    # @return [void]
+    #
+    def validate_license_info
+      # First check the project licensing information
+
+      # Check existance of licensing information
```
scotthain: :nitpick: `existence`

scotthain:
:+1:

lib/omnibus/licensing.rb
```diff
@@ -253,5 +297,99 @@ def local?(license)
...
+      "Zlib",          # zlib/libpng license
+      #
+      # In addition to these we would like to add some of the licenses that
+      # are frequently used in our depedencies.
+      #
+      "Public-Domain", # https://opensource.org/faq#public-domain
+      "Ruby",          # http://www.ruby-lang.org/en/LICENSE.txt
+      "Erlang-Public", # http://www.erlang.org/EPLICENSE
+      "Oracle-Binary", # http://www.oracle.com/technetwork/java/javase/terms/license/index.html
+      "Postgresql",    # http://www.postgresql.org/about/licence/
```
jamesc: This is already above - `PostgreSQL`

lib/omnibus/licensing.rb
```diff
@@ -253,5 +297,98 @@ def local?(license)
     def licensing_warning(message)
```
schisamo: What purpose does this special helper serve? In the rest of the code base we would just call `log.warn` directly.

sersut: In the future, I'm hoping to make it so that this helper checks a CLI flag and fails the run if set.

PR #640, opened 2016-03-17
Fix Windows builds

jkeiser:
This is necessary to get chef-dk using compiled ruby.

Two fixes:

1. Don't change newlines when `git pull`ing.
2. Remove the `sse` flags because we can't compile on virtualbox without them.

### DIFF
--- a/lib/omnibus/fetchers/git_fetcher.rb
+++ b/lib/omnibus/fetchers/git_fetcher.rb
@@ -197,13 +197,20 @@ module Omnibus
     #
     # Execute the given git command, inside the +project_dir+.
     #
+    # autcrlf is a hack to help support windows and posix clients using the
+    # same repository but canonicalizing files as they are committed to the
+    # repo but converting line endings when they are actually checked out
+    # into a working tree. We do not want to change the on-disk representation
+    # of our sources regardless of the platform we are building on unless
+    # explicitly asked for. Hence, we disable autocrlf.
+    #
     # @see Util#shellout!
     #
     # @return [Mixlib::ShellOut]
     #   the shellout object
     #
     def git(command)
-      shellout!("git #{command}", cwd: project_dir)
+      shellout!("git -c core.autocrlf=false #{command}", cwd: project_dir)
     end
 
     # Class methods
--- a/lib/omnibus/software.rb
+++ b/lib/omnibus/software.rb
@@ -614,8 +614,16 @@ module Omnibus
           opt_flag = windows_arch_i386? ? "-march=i686" : "-march=x86-64"
           {
             "LDFLAGS" => "-L#{install_dir}/embedded/lib #{arch_flag}",
-            # If we're happy with these flags, enable SSE for other platforms running x86 too.
-            "CFLAGS" => "-I#{install_dir}/embedded/include #{arch_flag} -O3 -mfpmath=sse -msse2 #{opt_flag}"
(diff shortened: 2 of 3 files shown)


lib/omnibus/fetchers/git_fetcher.rb
```diff
@@ -203,7 +203,7 @@ def contains_revision?(rev)
     #   the shellout object
     #
     def git(command)
-      shellout!("git #{command}", cwd: project_dir)
+      shellout!("git -c core.autocrlf=false #{command}", cwd: project_dir)
```
scotthain: I wonder if this belongs in https://github.com/chef-cookbooks/omnibus/blob/master/recipes/_git.rb rather than forcing it on every platform? I'm just worried about unintended consequences.

jkeiser: @scotthain omnibus feels like the right place because it's a requirement to build some things. Like, it would feel weird if the development environment had to have the right configuration when we already know what we need and can pass it.

I can imagine us doing it only on Windows, but it feels right to do it on all ... cc @ksubrama

danielsdeleo: AFAIK the crlf stuff only really applies to Windows, and for the kind of software we're building and the way we build it, we want it off everywhere.

lib/omnibus/software.rb
```diff
@@ -615,7 +615,7 @@ def with_standard_compiler_flags(env = {}, opts = {})
           {
             "LDFLAGS" => "-L#{install_dir}/embedded/lib #{arch_flag}",
             # If we're happy with these flags, enable SSE for other platforms running x86 too.
```
danielsdeleo: If we're removing SSE, then this comment probably doesn't make sense any longer.

adamedx: Yeah, should fix comment. :)

So this will affect chef-client, which was already released publicly with the sse stuff. We should just be sure this builds on a Windows system before merge.

schisamo:
Looks like the tests need updating.

lib/omnibus/fetchers/git_fetcher.rb
schisamo: @jkeiser Please add a comment above this line explaining why `core.autocrlf=false` is forced on every git call.

scotthain: @jkeiser I'm fine with that, just wanted to bring it up for discussion :) (I'm +1 on a comment as well)

ksubrama: Yeah I can add a comment to that.  autocrlf is a complete and utter hack (and so are it's bretheren in .gitattributes).  IMHO, a version control system should never be "guessing" the semantics of your file and modifying the syntax to suit.  That's the job of the compiler or IDE or whatever higher level tool whose job is to interpret your typing-thoughts.  autocrlf is a hack to allow a "canonical" representation of a file inside the underlying git repository while seamlessly checking it out with different line endings on different OSes.  That's what caused our build issue here - a certain config.hpp.in was being checked out on windows with \r\n line endings which makes an awk script not correctly match the line.  I figure that the source should look the same on whichever platform we build and if there is a translation needed, we should manually ask for it in the software definition (and this should happen regardless of whether we're on a build machine or on a developer build VM).

ksubrama:
:+1:

jkeiser:
:+1:

PR #641, opened 2016-03-18
skip files with whitespace on AIX/bff packaging

lamont-granquist:
This is the AIX version of #609

### DIFF
--- a/lib/omnibus/packagers/bff.rb
+++ b/lib/omnibus/packagers/bff.rb
@@ -133,7 +133,7 @@ module Omnibus
       # Get a list of all files
       files = FileSyncer.glob("#{staging_dir}/**/*").reject do |path|
         # remove any files with spaces.
-        if path =~ /\s+/
+        if path =~ /[[:space:]]/
           log.warn(log_key) { "Skipping packaging '#{path}' file due to whitespace in filename" }
           true
         end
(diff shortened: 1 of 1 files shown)


lamont-granquist:
@chef/omnibus-maintainers this is necessary to get ruby-2.2.4 building+installing on AIX.

i tried adding spaces to the regex that does the file-renaming dance, but couldn't make it work.  this works, and if it ever breaks builds then solaris will be even worse anyway...

lib/omnibus/packagers/bff.rb
```diff
@@ -131,8 +131,14 @@ def write_scripts
     def write_gen_template
 
       # Get a list of all files
-      files = FileSyncer.glob("#{staging_dir}/**/*").map do |path|
-
+      files = FileSyncer.glob("#{staging_dir}/**/*").reject do |path|
+        # remove any files with spaces.
+        if path =~ /\s+/
```
sethvargo: `[[:space:]]` will cover more things here

lamont-granquist: heh, i can't wait to see gems with checked in spec files that have unicode emojis in filenames and what those do to bff and solaris pkg packaging...

schisamo:
I'm :+1: if @scotthain is happy.

scotthain:
Ugh. Yep. That's a thing :+1:

PR #642, opened 2016-03-18
filter "all files" before ldd checking in health_check_ldd

srenatus:
This change to the "producer" of the `ldd`-check-pipe allows ignoring file names that are guaranteed to lead to non-matches.

I've noticed that our builds spend much of their time health-checking, but I haven't done any proper analysis. What do you think?

### DIFF
--- a/lib/omnibus/health_check.rb
+++ b/lib/omnibus/health_check.rb
@@ -151,6 +151,49 @@ module Omnibus
       /libmd\.so/,
     ].freeze
 
+    IGNORED_ENDINGS = %w(
+      .[ch]
+      .e*rb
+      .gemspec
+      .gitignore
+      .h*h
+      .java
+      .js
+      .json
+      .lock
+      .log
+      .lua
+      .md
+      .mkd
+      .out
+      .pl
+      .pm
+      .png
+      .py[oc]*
+      .r*html
+      .rdoc
+      .ri
+      .sh
+      .sql
+      .toml
+      .ttf
+      .txt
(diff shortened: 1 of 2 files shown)


lib/omnibus/health_check.rb
```diff
@@ -151,6 +151,16 @@ class HealthCheck
...
 
+    IGNORED_ENDINGS = %w(
+    Gemfile LICENSE VERSION README Rakefile .[ch] .rdoc .js .json .gemspec .e*rb
+    .ri .lock .md .yml .txt .png .pm .py[oc]* .sh .toml .gitignore .r*html .log
+    .out .pl .lua .h*h .sql .ttf .xml .mkd .java
+    ).freeze
+
+    IGNORED_PATTERNS = %w(
+    /terminfo/ /share/doc/ /share/postgresql/ /share/terminfo/
+    ).freeze
```
schisamo: Let's place one array item per line (and in alpha order) so adding additional items to `IGNORED_ENDINGS` or `IGNORED_PATTERNS` in the future produces an easier to review git change. For example:

```ruby
IGNORED_PATTERNS = %w(
  /share/doc/
  /share/postgresql/
  /share/terminfo/
  /terminfo/ 
).freeze
```

danielsdeleo:
:+1:

scotthain:
This is an awesome addition! :+1:

yzl:
:+1:

jaymalasinha:
:+1:

PR #643, opened 2016-03-18
Add `github` source that resolves to `git`

jkeiser:


lamont-granquist:
:+1:

ksubrama:
jkeiser is working on an update to ensure :override works as expected.

jkeiser:
@ksubrama @lamont-granquist updated to work with `:override`

lamont-granquist:
:+1:

lib/omnibus/software.rb
```diff
@@ -1077,15 +1080,26 @@ def path_key
...
+        override ||= {}
         val.merge(override)
       else
-        overrides[attr] || val
+        override || val
+      end
+    end
+
+    #
+    # Transform github -> git in source
```
sethvargo: Should make a note that this function modifies the given hash. I would suggest renaming to `canonicalize_source!` or duping the source hash before deleting a value.

sethvargo:
It should be fairly easy to unit test the `canonicalize_source` function.

lib/omnibus/software.rb
```diff
@@ -280,6 +280,8 @@ def source(val = NULL)
             "be a kind of `Hash', but was `#{val.class.inspect}'")
         end
 
+        val = canonicalize_source(val)
```
sethvargo: Need to update docs above to include `:github` as an option

lib/omnibus/software.rb
```diff
@@ -1077,15 +1080,26 @@ def path_key
...
+        override || val
+      end
+    end
+
+    #
+    # Transform github -> git in source
+    #
+    def canonicalize_source(source)
+      if source.is_a?(Hash) && source[:github]
+        source[:git] = "https://github.com/#{source[:github]}.git"
```
sethvargo: It might be worth noting (somewhere) that github URLs will use the git-over-https and not git+ssh, so users will need to have their keychain setup correctly to pull.

ksubrama:
:+1:

schisamo:
Let's be sure to address all of @sethvargo's review feedback before a merge.

jkeiser:
@schisamo @sethvargo yep! Soon as I get a free cycle I'll get on it, thanks :)

jkeiser:
@schisamo @sethvargo all comments addressed, thanks for the thorough review :)

lib/omnibus/software.rb
sethvargo: Hey @jkeiser it looks like my other comment got lost in a commit, but this is still a bit confusing. The _return_ value of the `canonicalize_source` function is a hash, but the function itself also modifies the given hash in place (`Hash#delete` is a destructive operation). So you're modifying the return hash, then setting it to the value of self. Not only is that a weird reflective operation, but it could be very confusing to users. The "best practice" is to either make this a bang method without reassignment

```ruby
canonicalize_source!(val)
```

or rewrite the method to dup the given hash:

```ruby
val = canonicalize_source(val)
# ...
def canonical_source(hash)
  hash = hash.dup
end
```

jkeiser: canonicalize_source was changed to `dup` the hash, yep.

jkeiser: @sethvargo I put a `dup` in here: https://github.com/chef/omnibus/blob/master/lib/omnibus/software.rb#L1101

Sorry to respond after merge, I didn't see your comment because it appeared way up above the fold. Does this resolve what you were after? If not, I'll happily make a new PR!

jkeiser:
FYI this is merged. I'm unclear why github didn't recognize it. Again @sethvargo, sorry I missed your late comment, but I believe it's addressed. Let me know if you're still concerned and I'll get a new PR up ASAP. Thanks for the thorough review!

PR #644, opened 2016-03-18
WIP - Remove solaris mapfile for Solaris11 and above

jaymalasinha:
Solaris 11 and above does not require the custom library mappings.
Tested the update successfully on Solaris11 VM
@chef/engineering-services

### DIFF
--- a/lib/omnibus/software.rb
+++ b/lib/omnibus/software.rb
@@ -655,7 +655,7 @@ module Omnibus
         "LD_RUN_PATH" => "#{install_dir}/embedded/lib"
       }
 
-      if solaris2?
+      if solaris2? && Ohai['os_version'].to_i <= 5.10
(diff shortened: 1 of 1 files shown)


scotthain:
:+1:  pending tests passing

lib/omnibus/software.rb
```diff
@@ -655,7 +655,7 @@ def with_standard_compiler_flags(env = {}, opts = {})
         "LD_RUN_PATH" => "#{install_dir}/embedded/lib"
       }
 
-      if solaris2?
+      if solaris2? && Ohai['os_version'].to_i <= 5.10
```
schisamo: We should use an actual version constraint comparison here, something like this should work:

```ruby
if solaris2? && platform_version.satisfies?('<= 5.10')
```

scotthain:
This was resolved by https://github.com/chef/omnibus/pull/650

PR #645, opened 2016-03-18
Allow users to generate a manifest for other platforms

jkeiser:
We are interested in generating manifests for chef and chef-dk tentatively, and checking them in to the repository. This patch allows us to do that from the command line for all OS's we support.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@ gemspec
 
 # Fork to allow for a recent version of multipart-post.
 gem 'pedump', git: 'https://github.com/ksubrama/pedump', branch: 'patch-1'
+gem 'pry'
 
 group :docs do
   gem 'yard',          '~> 0.8'
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,6 @@ gemspec
 
 # Fork to allow for a recent version of multipart-post.
 gem 'pedump', git: 'https://github.com/ksubrama/pedump', branch: 'patch-1'
-gem 'pry'
 
 group :docs do
   gem 'yard',          '~> 0.8'
--- a/features/step_definitions/generator_steps.rb
+++ b/features/step_definitions/generator_steps.rb
@@ -1,7 +1,6 @@
 require 'aruba/api'
 
 Given(/^I have an omnibus project named "(.+)"$/) do |name|
-  # Make an omnibus config per project
   create_directory(name)
   cd(name)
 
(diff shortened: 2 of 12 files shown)


schisamo:
@jkeiser Now that `omnibus manifest` has numerous CLI options we should add some Aruba coverage here:
https://github.com/chef/omnibus/tree/master/features/commands

yzl:
:+1: pending Aruba coverage.

jkeiser:
@yzl @schisamo I added aruba coverage. I had to modify `Omnibus.reset!` to reset cached project and software definitions to make it work, but it's in good shape now, I think. Also added a couple of build tests.

Gemfile
```diff
@@ -3,6 +3,7 @@ gemspec
 
 # Fork to allow for a recent version of multipart-post.
 gem 'pedump', git: 'https://github.com/ksubrama/pedump', branch: 'patch-1'
+gem 'pry'
```
yzl: Do we want `pry` all the time, or is this for debugging?

features/step_definitions/generator_steps.rb
```diff
@@ -28,6 +35,11 @@
   EOH
 end
 
+Given(/^I debug$/) do
```
yzl: I'd rather not add this if it means adding `pry` as a dep, but will defer to others' opinion.

scotthain: I'd also like to avoid `pry` as a dep if at all possible.

Gemfile
jkeiser: This is for debugging only (thus putting in the gemfile so people don't get it when they `gem install omnibus`).

features/step_definitions/generator_steps.rb
jkeiser: `pry` isn't a dep, it's only in the `Gemfile`. `gem install omnibus` won't install it.

schisamo: @jkeiser Omnibus isn't installed that way though...it's pulled in via a `bundle install` from an Omnibus project. Either way we should place `pry` in a `:development` group in the `Gemfile` OR probably just make it a development dependency in the gemspec:

```ruby
gem.add_development_dependency 'pry'
```

Gemfile
sethvargo: It'll actually get pulled in unless you wrap it in `group :development`

features/step_definitions/generator_steps.rb
sethvargo: The advantage to putting it in the gemspec as a dev dep as opposed to in the gemfile is that it shows up on rubygems.org as a dev dep (gemfile deps dont, but still get pulled in)

Gemfile
jkeiser: And in goes a `group :development`!

features/step_definitions/generator_steps.rb
jkeiser: Will do!

jkeiser:
@schisamo @yzl @sethvargo development dependency fixed.

lib/omnibus/project.rb
```diff
@@ -49,6 +49,10 @@ def load(name, manifest=nil)
         end
       end
 
+      def reset!
```
schisamo: Need a yard comment here, see the following for a good example:
https://github.com/chef/omnibus/blob/798b2db8e34a296d074006086288cfb6693c9d65/lib/omnibus/config.rb#L76-L81

lib/omnibus/software.rb
```diff
@@ -52,6 +52,10 @@ def load(project, name, manifest)
         end
       end
 
+      def reset!
```
schisamo: Need a yard comment here, see the following for a good example:
https://github.com/chef/omnibus/blob/798b2db8e34a296d074006086288cfb6693c9d65/lib/omnibus/config.rb#L76-L81

lib/omnibus/project.rb
jkeiser: Done!

schisamo:
@jkeiser Let's squash this down into 1 or 2 commits please (and ensure new feature commits have a descriptive message about the feature). Otherwise I am :+1:

scotthain:
:+1: with squash

jkeiser:
Squashed!

schisamo:
@jkeiser MWR once Travis and AppVeyor checks are green.

PR #646, opened 2016-03-19
Get DEB architecture from the packager directly

edolnx:
We currently use `uname -m` to infer the architecture for a system, but this requires a lot of extra logic and doesn't always match 1:1 with a distro. For example: in the world of Debian gcc and kernel architecture names do not match Debian architecture names. It's further complicated in the world of ARM where the runtime architecture (armhf or armel) are designed to span processors architectures (ARMv6, ARMv7. and ARMv8).

This resolves all of that by getting the native architecture of the runtime environment from the native packaging tools in DEB environments by getting the system architecture directly from `dpkg`. This works for omnibus, but isn't a good fit for a ohai plugin becuase while `dpkg` is in all Debain runtime environments (or should be), there does not appear to be a reliable way to detect this properly in an RPM environment using the packaging tools.

Many tests are removed which were just ensuring that the platform we computed matched what was expected. The `safe_architecture` functions in the respective libraries also fall back to a known default to allow tests to succeed when the tool is not available (ex: rpm test on a deb system).

This code was tested on several architectures and distributions.

### DIFF
--- a/lib/omnibus/packagers/deb.rb
+++ b/lib/omnibus/packagers/deb.rb
@@ -397,29 +397,7 @@ module Omnibus
     # @return [String]
     #
     def safe_architecture
-      case Ohai['kernel']['machine']
-      when 'x86_64'
-        'amd64'
-      when 'i686'
-        'i386'
-      when /armv\dl/
-        if Ohai['platform'] == 'raspbian' || Ohai['platform'] == 'ubuntu'
-          'armhf'
-        else
-          Ohai['kernel']['machine']
-        end
-      when 'aarch64'
-        # Debian prefers amd64 on ARMv8/AArch64 (64bit ARM) platforms
-        # see https://wiki.debian.org/Arm64Port
-        'arm64'
-      when 'ppc64le'
-        # Debian prefers to use ppc64el for little endian architecture name
-        # where as others like gnutools/rhel use ppc64le( note the last 2 chars)
-        # see http://linux.debian.ports.powerpc.narkive.com/8eeWSBtZ/switching-ppc64el-port-name-to-ppc64le
-        'ppc64el'  #dpkg --print-architecture = ppc64el
-      else
-        Ohai['kernel']['machine']
-      end
+      @safe_architecture ||= shellout!("dpkg --print-architecture").stdout.split("\n").first || "noarch"
     end
   end
 end
--- a/spec/unit/packagers/deb_spec.rb
+++ b/spec/unit/packagers/deb_spec.rb
@@ -107,6 +107,10 @@ module Omnibus
     end
 
     describe '#write_control_file' do
+      before do
+        allow(subject).to receive(:safe_architecture).and_return("amd64")
+      end
+
       it 'generates the file' do
         subject.write_control_file
         expect("#{staging_dir}/DEBIAN/control").to be_a_file
@@ -224,6 +228,7 @@ module Omnibus
       before do
         allow(subject).to receive(:shellout!)
         allow(Dir).to receive(:chdir) { |_, &b| b.call }
+        allow(subject).to receive(:safe_architecture).and_return("amd64")
       end
 
       it 'logs a message' do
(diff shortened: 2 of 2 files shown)


edolnx:
cc @chef/engineering-services @chef/omnibus-maintainers

lib/omnibus/packagers/deb.rb
```diff
@@ -397,28 +397,11 @@ def safe_version
...
-        # Debian prefers to use ppc64el for little endian architecture name
-        # where as others like gnutools/rhel use ppc64le( note the last 2 chars)
-        # see http://linux.debian.ports.powerpc.narkive.com/8eeWSBtZ/switching-ppc64el-port-name-to-ppc64le
-        'ppc64el'  #dpkg --print-architecture = ppc64el
-      else
-        Ohai['kernel']['machine']
+      begin
+        so = shell_out("dpkg --print-architecture")
+        return so.stdout.split($/) || "noarch"
+      rescue
```
schisamo: Please specify the exception class you are hoping to catch so we don't accidentally suppress other error output. 

If this rescue is here simply for test suites:

* Mock out `Omnibus::Packagers::Deb#safe_architecture` by default.
* Any tests for `Omnibus::Packagers::Deb#safe_architecture` should mock out `shell_out`.

lib/omnibus/packagers/deb.rb
```diff
@@ -397,28 +397,11 @@ def safe_version
...
-        'arm64'
-      when 'ppc64le'
-        # Debian prefers to use ppc64el for little endian architecture name
-        # where as others like gnutools/rhel use ppc64le( note the last 2 chars)
-        # see http://linux.debian.ports.powerpc.narkive.com/8eeWSBtZ/switching-ppc64el-port-name-to-ppc64le
-        'ppc64el'  #dpkg --print-architecture = ppc64el
-      else
-        Ohai['kernel']['machine']
+      begin
+        so = shell_out("dpkg --print-architecture")
```
schisamo: We want the bang (`!`) of here...so `shell_out!`.

lib/omnibus/packagers/rpm.rb
```diff
@@ -559,17 +559,11 @@ def safe_version
...
-          'armv6hl'
-        else
-          'armv6l'
-        end
-      else
-        Ohai['kernel']['machine']
+      begin
+        so = shell_out("rpmspec -E \"%{_arch}\"")
+        return so.stdout.split($/) || "noarch"
+      rescue
```
schisamo: Similar comment as I left on `Omnibus::Packagers::Deb`: https://github.com/chef/omnibus/pull/646/files#r56850540

schisamo:
@edolnx Left a few comments. What are the actual 32-bit and 64-bit architecture string values this will begin returning on Ubuntu and EL?

edolnx:
I just completed testing on various platforms and architectures, tests pass and packages build as expected! (better news: unit tests for omnibus now work on non-x86 platforms)

yzl:
@edolnx This looks good.  Could you add a note to the PR about the platforms/architectures on which you performed testing?

edolnx:
Sure thing:

Ubuntu 15.10 amd64
Centos 7 x86_64
Debian 8 amd64
Debian 8 arm64
Debian 8 armhf
Debian 8 armel

I can test others if desired, like ppc64 and ppc64le which will be done in a few days once I get those VMs talking again. Also hoping to test on Fedora armv7 and armv8 soon.

scotthain:
:+1: can we send a quick angrychef build through CI to be super safe and verify we don't have any surprises on the 'edge case' platforms that use rpm/deb?

edolnx:
The angrychef tests show that there is no reliable way to detect RPM platform correctly other than what we were previously doing. Reverted that code. The DEB stuff is rock solid, and makes that code much simpler. The `rpmspec` command is only available in EL7 (and possibly later), and there is no equivalent command for previous releases.

schisamo:
@edolnx Thanks for the testing, let's squash this down into a single commit and ensure that commit has a well formed message.

edolnx:
Squashed and ready to go

schisamo:
@chef/omnibus-maintainers please give this final review.

scotthain:
Love the simplification of the tests.

lib/omnibus/packagers/deb.rb
```diff
@@ -397,29 +397,8 @@ def safe_version
...
-      when 'ppc64le'
-        # Debian prefers to use ppc64el for little endian architecture name
-        # where as others like gnutools/rhel use ppc64le( note the last 2 chars)
-        # see http://linux.debian.ports.powerpc.narkive.com/8eeWSBtZ/switching-ppc64el-port-name-to-ppc64le
-        'ppc64el'  #dpkg --print-architecture = ppc64el
-      else
-        Ohai['kernel']['machine']
-      end
+      so = shellout!("dpkg --print-architecture")
+      return so.stdout.split("\n").first || "noarch"
```
schisamo: We should cache the result here (so we don't shell out each time `safe_architecture` is called). Something like this should work:

```ruby
@safe_architecture ||= begin
  so = shellout!("dpkg --print-architecture")
  so.stdout.split("\n").first || "noarch"
end
```

sethvargo: Better yet, since we know we are on debian, we can leverage unix tools:

```
@safe_architecture ||= shellout!("dpkg --print-architecture | head -n1").strip || "noarch"
```

edolnx: I'll add the caching, but putting in a full unix pipeline means errors we expect to cause an exception will be silently caught and lead to unpredictable results.

PR #653, opened 2016-03-28
Use project.license in deb and rpm packagers if available.

sersut:
### Description

This PR uses the `license` set in the project if available in rpm / deb packagers so that users can find the license information a bit more easy. 

/cc @chef/omnibus-maintainers @jamesc


jamesc:
+1

PR #655, opened 2016-04-01
Pin jmespath to 1.1.x: 1.2.2 brings in json_pure and breaks us

jkeiser:
### Fix issue with json_pure killing chef+chef-dk builds

aws-sdk brings in jmespath, and jmespath 1.2.2 brings in json_pure, which breaks with this error:

```
/home/jenkins/workspace/chefdk-build/architecture/x86_64/platform/ubuntu-12.04/project/chefdk/role/builder/omnibus/vendor/bundle/ruby/2.1.0/gems/json_pure-1.8.3/lib/json/pure/generator.rb:366:in `to_json': wrong argument type JSON::Pure::Generator::State (expected Data) (TypeError)`
```

/cc @chef/omnibus-maintainers

### DIFF
--- a/omnibus.gemspec
+++ b/omnibus.gemspec
@@ -31,6 +31,8 @@ Gem::Specification.new do |gem|
   gem.add_dependency 'ohai',             '~> 8.0'
   gem.add_dependency 'ruby-progressbar', '~> 1.7'
   gem.add_dependency 'aws-sdk',          '~> 2'
+  # aws-sdk brings this in, and jmespath 1.2+ brings in json_pure, which fails the build
+  gem.add_dependency 'jmespath',         '< 1.2'
   gem.add_dependency 'thor',             '~> 0.18'
 
   gem.add_development_dependency 'bundler'


schisamo:
:+1: Assuming it passes all testing (including a full integration build on Manhattan).

yzl:
:+1: to pinning `jmespath`.

schisamo:
I would rather we don't do transitive pinnings. I'm fine with pining to a certain version of `aws-sdk` but pinning to transitive deps in a framework library makes my skin crawl.

We should also fix the underlying JSON issue.

lamont-granquist:
who is calling #to_json and sending in arguments that make json_pure barf?

danielsdeleo:
I'm :+1: I don't mind transitive pinning so much, as long as it's clearly explained what's going on and when we can remove it.

scotthain:
Has the underlying code been identified at least? I am ok with this as a purely temporary fix, but I'd love to see a link to a PR or issue that resolves the actually breakage.

jkeiser:
I'm closing this; we'll figure it out elsewhere.

PR #657, opened 2016-04-05
Manifest Fixes/Improvements

schisamo:
### Description

* Add proper support for loading v2 manifests - This change was overlooked in #635.
* Use symbolized keys for all Manifest hashes - This ensures are internal Hash representations are consistent across objects (e.g. Metadata).

### DIFF
--- a/spec/unit/software_spec.rb
+++ b/spec/unit/software_spec.rb
@@ -505,7 +505,7 @@ module Omnibus
       let(:manifest_entry) {Omnibus::ManifestEntry.new("software", {locked_version: "1.2.8", locked_source: a_source})}
       let(:manifest) do
         m = Omnibus::Manifest.new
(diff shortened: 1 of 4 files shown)


spec/unit/software_spec.rb
```diff
@@ -505,7 +505,7 @@ module Omnibus
       let(:manifest_entry) {Omnibus::ManifestEntry.new("software", {locked_version: "1.2.8", locked_source: a_source})}
```
scotthain: Would adding a test that utilizes a `ManifestEntry` that includes a license be of any value? I'm guessing not...

schisamo: I think we are OK without as the license is just part of the 2nd argument hash.

scotthain: That's what I figured. :shipit:

PR #658, opened 2016-04-05
Don't publish full version manifest to Artifactory

schisamo:
Publishing a package with a large version manifest fails with the following cryptic error:

```
/opt/languages/ruby/2.1.5/lib/ruby/2.1.0/net/http/generic_request.rb:202:in
`copy_stream': Broken pipe - sendfile (Errno::EPIPE)
```

The error is related to the way in which properties are set via Artifactory's REST API. The properties are attached to the upload URI as matrix parameters:
https://www.jfrog.com/confluence/display/RTF3X/Artifactory+REST+API#ArtifactoryRESTAPI-SetItemProperties
https://www.w3.org/DesignIssues/MatrixURIs.html

We'll avoid this issue by publishing a subset of the version manifest data as the `omnibus.dependencies` property. This lighterweight hash will only contain a list of dependencies and their associated version and license data.

/cc @chef/omnibus-maintainers 

:construction: I will perform some end-to-end integration testing with this updated publisher before merging.

### DIFF
--- a/lib/omnibus/publishers/artifactory_publisher.rb
+++ b/lib/omnibus/publishers/artifactory_publisher.rb
@@ -205,7 +205,7 @@ module Omnibus
         'omnibus.sha1'             => package.metadata[:sha1],
         'omnibus.sha256'           => package.metadata[:sha256],
         'omnibus.sha512'           => package.metadata[:sha512],
-        'omnibus.version_manifest' => package.metadata[:version_manifest] || '',
+        'omnibus.dependencies'     => dependency_data_for(package),
       }.tap do |h|
         if build_record?
           h['build.name'] = package.metadata[:name]
@@ -256,6 +256,28 @@ module Omnibus
         package.metadata[:platform_version],
         package.metadata[:basename],
       )
+    end
+
+    #
+    # Dependency data for all of a package
+    #
+    # @param [Package] package
+    #   the package to generate the dependency data for
+    #
+    # @return [Hash<String, String>]
+    #
+    def dependency_data_for(package)
+      dependency_data = {}
+      dep_list = package.metadata.to_hash.fetch(:version_manifest, {}).fetch(:software, {})
+      dep_list.each do |name, data|
+        dependency_data[name.to_s] = {
+          'source_type'       => data[:source_type],
+          'described_version' => data[:described_version],
+          'locked_version'    => data[:locked_version],
+          'license'           => data[:license],
+        }
+      end
+      dependency_data
     end
   end
 end
--- a/spec/unit/publisher_spec.rb
+++ b/spec/unit/publisher_spec.rb
@@ -74,18 +74,64 @@ module Omnibus
             sha1: 'SHA1',
             md5: 'ABCDEF123456',
             version_manifest: {
-              'manifest_format' => 1,
-              'build_version' => '11.0.6',
-              'build_git_revision' => '2e763ac957b308ba95cef256c2491a5a55a163cc',
-              'software' => {
-                'zlib' => {
-                  'locked_source' => {
-                    'url' => 'an_url'
+              manifest_format: 1,
+              build_version: '11.0.6',
+              build_git_revision: '2e763ac957b308ba95cef256c2491a5a55a163cc',
+              software: {
+                zlib: {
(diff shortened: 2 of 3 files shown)


scotthain:
Looks solid, 👍 pending successful e2e tests.

schisamo:
Artifactory properties attached to a published package now look like (note `omnibus.dependencies`):

artifactory_chef_s_preprod_artifactory_instance____repository_browser

@josephrdsmith @charlesjohnson Does the `omnibus.dependencies` value contain enough data to complete the _Open Source License Management_ feature? I'm a little sketchy on exactly what will need to be displayed on downloads.chef.io to fulfill the requirement.

sersut:
@schisamo From our previous discussions with @jamesc on "What do the users want to see in terms of licenses?" generally the answer is "They want to see the contents of all the licenses." 

So IMO if we are going to show users some license information when they are downloading the package, it should be the actual contents of its licenses. I would suggest that we add a new field there that contains the content of the project's license if it exists. 

I.e. 

```
omnibus.licenses = IO.read(project.license_file_path)
```

The contents of that file will include the text of the project's license as well as version and license name of the dependencies. I think with that we will have pretty good coverage.

My .02

schisamo:
@sersut The contents of the license will be too much data to shove in an Artifactory property, these properties should be thought of as tags in other systems. The best we can do is use the values in `omnibus.license` (project license) and each dep's (in `omnibus.dependencies`) license value as a pointer to the full text.

schisamo:
Just to ensure this conversation is updated with what was discussed in Slack, we need enough data to construct something similar to:
https://www.chef.io/3rd-party-licenses/

I believe the license name will be enough data from this layer of the feature implementation.

jamesc:
@schisamo What we really need is to ship the LICENSE file that is in the root directory.  That gives everything they need (which is the pointer to where we got the various pieces of code we used).  This would eventually go onto downloads.chef.io.

schisamo:
@jamesc That is much easier said than done. The contents of the `LICENSE` file will be much too large to shove in an Artifactory property (e.g. `omnibus.licenses`). Alternatively we try and flow a package's associated `LICENSE` file (or its `*.metadata.json` file) from the build system to **artifactory.chef.co** to **packages.chef.io/stable**. This will require a lot of changes in our release infra.

Can we take an simpler approach similar to https://www.chef.io/3rd-party-licenses/ and still meet our requirements?

cwebberOps:
@schisamo does it make sense to just have an S3 bucket that we publish a copy of the file to? I know it is kinda jank, but it becomes an easy way to always be able to reference.

schisamo:
@cwebberOps @jamesc So the generated `LICENSE` file in one of our packages doesn't contain any more extended license data than the `omnibus.dependencies` property in my screenshot above. It's really just a list of deps, their version, the name of the license the dep uses and the full path on disk to their individual license files (vs the actual content).  For example, here is the `/opt/chef/LICENSE` from a Chef build: https://gist.github.com/schisamo/ea7d33f35314017d9dbd35ca669729a0

As for the actual license of a product (vs the deps) we only use two licenses: `Apache-2.0` and `Chef-MSLA`. It would be easy to have the contents of these two licenses in the chef-web-downloads source code, or we can just link to a definitive (and always up to date) copy somewhere else:
http://www.apache.org/licenses/LICENSE-2.0
https://www.chef.io/online-master-agreement/

I think for a v1 of this feature the content of the `omnibus.dependencies` and `omnibus.license` properties should be enough to build out the UI. This data can easily be pulled from artifactory.chef.co and appended to the product data yaml when a `rake fetch` is run in chef-web-downloads.

jamesc:
@schisamo yeah, I agree. I was thinking that last might that we could just regen the LICENSE ile contents on the download site

schisamo:
I did some test publishing using these updates and some projects produce an `omnibus.dependencies` value that is too large (and thus publishing fails). I think the sanest way forward is to begin publishing a package's `*.metadata.json` file alongside it.  This will allow us to shove any data we desire into this JSON file and easily retrieve it when downloads.chef.io is re-generated. I'll work up a new PR early next week for further discussion. Closing this PR for now though.

Issue #660, opened 2016-04-07
Only warn when license_file for the project is missing


sersut:
When the `license_file` pointed to by the project does not exist, omnibus build should not fail. It should only log a warning message.

We ran into this recently in chef and chefdk builds.

```
                        [Licensing] W | Software 'config_guess' uses license 'GPL-3.0 (with exception)' which is not one of the standard licenses identified in https://opensource.org/licenses/alphabetical. Consider using one of the standard licenses.
                        [Licensing] W | License file '/var/cache/omnibus/src/rubygems/LICENSE.txt' does not exist for software 'rubygems'.
                        [Licensing] W | License file '/var/cache/omnibus/src/pry/LICENSE' does not exist for software 'pry'.
/usr/home/jenkins/workspace/chef-build/architecture/x86_64/platform/freebsd-10/project/chef/role/builder/omnibus/vendor/bundle/ruby/2.1.0/bundler/gems/omnibus-cdd633075fb6/lib/omnibus/licensing.rb:179:in `read': No such file or directory @ rb_sysopen - /usr/home/jenkins/workspace/chef-build/architecture/x86_64/platform/freebsd-10/project/chef/role/builder/omnibus/LICENSE (Errno::ENOENT)
	from /usr/home/jenkins/workspace/chef-build/architecture/x86_64/platform/freebsd-10/project/chef/role/builder/omnibus/vendor/bundle/ruby/2.1.0/bundler/gems/omnibus-cdd633075fb6/lib/omnibus/licensing.rb:179:in `project_license_content'
	from /usr/home/jenkins/workspace/chef-build/architecture/x86_64/platform/freebsd-10/project/chef/role/builder/omnibus/vendor/bundle/ruby/2.1.0/bundler/gems/omnibus-cdd633075fb6/lib/omnibus/licensing.rb:129:in `block in create_project_license_file'
	from /usr/home/jenkins/workspace/chef-build/architecture/x86_64/platform/freebsd-10/project/chef/role/builder/omnibus/vendor/bundle/ruby/2.1.0/bundler/gems/omnibus-cdd633075fb6/lib/omnibus/licensing.rb:126:in `open'
	from /usr/home/jenkins/workspace/chef-build/architecture/x86_64/platform/freebsd-10/project/chef/role/builder/omnibus/vendor/bundle/ruby/2.1.0/bundler/gems/omnibus-cdd633075fb6/lib/omnibus/licensing.rb:126:in `create_project_license_file'
[...]

PR #662, opened 2016-04-08
Warn when the specified license_file for the project does not exist

sersut:
### Description

This PR fixes https://github.com/chef/omnibus/issues/660. We ran into this in chef and chefdk yesterday when the license_file for the project was specified incorrectly. In general we never want to fail due to licensing information for now. 

/cc @chef/omnibus-maintainers, @patrick-wright

### DIFF
--- a/lib/omnibus/licensing.rb
+++ b/lib/omnibus/licensing.rb
@@ -176,7 +176,15 @@ module Omnibus
     # @return [String]
     #
     def project_license_content
-      project.license_file.nil? ? "" : IO.read(File.join(Config.project_root,project.license_file))
+      return "" if project.license_file.nil?
+
+      full_license_file_path = File.join(Config.project_root,project.license_file)
+
+      if File.exist?(full_license_file_path)
+        IO.read(full_license_file_path)
+      else
+        licensing_warning("Specified license file '#{full_license_file_path}' does not exist for the project.")
+      end
     end
 
     #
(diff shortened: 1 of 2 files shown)


sethvargo:
Hey @sersut 

I think this is a bad idea. The current behavior is to do nothing if the license is not specified, but fail if the license is specified and does not exist. I think that's the best UX here. If you tell a machine to pull the license file from a place, it should fail if unable to do so. 

For example, in your case, if this was just a warning running in CI, would you have detected the missing license file?

sersut:
Good point @sethvargo. I wanted to do this because in general I did not want to fail the build due to incorrect licensing information (for now - until we get a `--fatal-licensing-errors` flag). 

But your point also makes sense. 

I will leave this open for others to chime in but I'm OK with not doing this.

lamont-granquist:
the ability to --fatal-licensing-errors (including both cant-find-it and you-didnt-tell-me-it errors) would probably be best...

schisamo:
I agree with @sethvargo on this one. Not being able to pull a specified license file is a failure in the build. It's a similar case as failing a health check IMO.

stevendanna:
I'd also like this to stay a hard-fail in the default case.  Failing a build because of a mis-spelling is indeed annoying for sure; but presumably if you have take the time to specify a license then you care deeply about shipping builds with correct licensing information and never shipping a build without correct information because of your mis-spelling would be worth that annoyance.

Overall the goal of these type of features in both omnibus and omnibus-software should be to serve as on-by-default verification that a package is built to the correct standard.

sersut:
:+1:

PR #664, opened 2016-04-11
Publish a package’s `*.metadata.json` to Artifactory

schisamo:
Publishing a package with a large `omnibus.version_manifest` property fails with the following cryptic error:

```
/opt/languages/ruby/2.1.5/lib/ruby/2.1.0/net/http/generic_request.rb:202:in
`copy_stream': Broken pipe - sendfile (Errno::EPIPE)
```

The error is caused by large property values and is related to the way in which properties are set in Artifactory's REST API. The properties are attached to the upload URI as matrix parameters:
https://www.jfrog.com/confluence/display/RTF3X/Artifactory+REST+API#ArtifactoryRESTAPI-SetItemProperties
https://www.w3.org/DesignIssues/MatrixURIs.html

We’ll workaround this issue by by publishing all `*.metadata.json` files alongside their respective packages. This will allow us to index additional information about a package or build while avoiding any technical limitations that Artifactory properties might have.

I've also added the actual content of a project's actual `LICENSE` file to the Metadata which should make retrieval/display easy on distribution services like downloads.chef.io.

/cc @chef/omnibus-maintainers @chef/engineering-services @sersut @jamesc @cwebberOps

### DIFF
--- a/lib/omnibus/metadata.rb
+++ b/lib/omnibus/metadata.rb
@@ -63,6 +63,7 @@ module Omnibus
           iteration:        project.build_iteration,
           license:          project.license,
           version_manifest: project.built_manifest.to_hash,
+          license_content:  File.exist?(project.license_file_path) ? File.read(project.license_file_path) : ''
         }
 
         instance = new(package, data)
--- a/lib/omnibus/publishers/artifactory_publisher.rb
+++ b/lib/omnibus/publishers/artifactory_publisher.rb
@@ -16,7 +16,6 @@
 
 require 'uri'
 require 'benchmark'
-require 'ffi_yajl'
 
 module Omnibus
   class ArtifactoryPublisher < Publisher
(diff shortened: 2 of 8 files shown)


lib/omnibus/publishers/artifactory_publisher.rb
```diff
@@ -29,17 +28,28 @@ def publish(&block)
...
+              properties,
+            )
+            # Upload the package's assoacited `*.metadata.json` file
+            log.info(log_key) { "Uploading '#{package.name}.metadata.json'" }
+            artifact_for(package.metadata).upload(
+              repository,
+              "#{remote_path}.metadata.json",
+              # *.metadata.json files should not include
+              # the checksum properties
+              properties.delete_if {|k,v| k =~ /^omnibus\.(md5|sha)/ },
```
sethvargo: Missing space

sethvargo: Just a note that `delete_if` will modify the in-place hash

sersut:
Codewise :+1:

lib/omnibus/metadata.rb
```diff
@@ -63,6 +63,7 @@ def generate(path, project)
           iteration:        project.build_iteration,
           license:          project.license,
           version_manifest: project.built_manifest.to_hash,
+          license_content:  File.read(project.license_file_path),
```
sersut: We probably want to project this with `File.exist?()` incase the project does not have this file present.

scotthain:
Code lgtm 👍

yzl:
:+1: pending integration testing

Issue #663, opened 2016-04-08
Omnibus RPM Owns Directory in Filesystem Package


eoly:
### Description

RPMs built by Omnibus are claiming ownership of directories owned by the filesystem package resulting in conflicts when installing the RPM on a Centos 7 system.

I think this may be a regression back to the behavior of Omnibus prior to PR #392 

It appears the regression was introduced in this commit: https://github.com/chef/omnibus/commit/160389f61a5758970f36df3a379551847cd1a79f
### Omnibus Version

omnibus (5.3.0)
### Platform Version

CentOS Linux release 7.2.1511 (Core)
### Replication Case

**Perform Steps on CentOS 7 VM or Docker Container**

Build Test Omnibus Package
- git clone https://github.com/eoly/omnibus-test
- cd omnibus-test
- mkdir pkg
- bundle install --binstubs
- bin/omnibus build test

Run rpm -qpl on the resulting RPM and you will see the RPM claims ownership of the /lib directory.

```
[root@localhost pkg]# rpm -qpl test-0.0.0+20160408161743-1.el7.x86_64.rpm
/lib
/lib/systemd
/lib/systemd/system
/lib/systemd/system/thing.service
/opt
/opt/test
/opt/test/LICENSE
/opt/test/LICENSES
/opt/test/bin
/opt/test/embedded
/opt/test/embedded/bin
/opt/test/embedded/lib
/opt/test/version-manifest.json
/opt/test/version-manifest.txt
```

Trying installing the rpm with rpm -ivh

```
[root@localhost pkg]# rpm -ivh test-0.0.0+20160408161743-1.el7.x86_64.rpm
Preparing...                          ################################# [100%]
    file /lib from install of test-0.0.0+20160408161743-1.el7.x86_64 conflicts with file from package filesystem-3.2-20.el7.x86_64
```

---

/cc @chef/omnibus-maintainers <- This ensures the Omnibus Maintainers team are notified to review this PR.

schisamo:
It sounds like we need to update:
https://github.com/chef/omnibus/blob/master/resources/rpm/filesystem_list

Can you run the following command and share the output:

```
rpm -ql filesystem-3.2-20.el7.x86_64
```

Feel free to just open a PR with the required changes.

eoly:
Updating filesystem_list based on the output of `rpm -ql filesystem-3.2-20.el7.x86_64` did not resolve the issue. I created PR #665 just in case you want to update the list anyways.

[...]

PR #666, opened 2016-04-12
exclude directories owned by the filesystem package

eoly:
### Description

This addresses #663.

Directories owned by the filesystem package should be excluded from the RPM produced by Omnibus. 

From: https://fedoraproject.org/wiki/Packaging:Guidelines#File_and_Directory_Ownership

> Directory ownership is a little more complex than file ownership. Packages must own all directories they put files in, except for:
> any directories owned by the filesystem, man, or other explicitly created -filesystem packages
> any directories owned by other packages in your package's natural dependency chain 

### DIFF
--- a/lib/omnibus/packagers/rpm.rb
+++ b/lib/omnibus/packagers/rpm.rb
@@ -391,12 +391,12 @@ module Omnibus
 
     #
     # Convert the path of a file in the staging directory to an entry for use in the spec file.
-    #
+    # Exclude directories owned by the filesystem package.
     # @return [String]
     #
     def build_filepath(path)
       filepath = rpm_safe('/' + path.gsub("#{build_dir}/", ''))
-      return if config_files.include?(filepath)
+      return if config_files.include?(filepath) || filesystem_directories.include?(filepath)
       full_path = build_dir + filepath.gsub('[%]','%')
       # FileSyncer.glob quotes pathnames that contain spaces, which is a problem on el7
       full_path.gsub!('"', '')


yzl:
Hello @eoly, thanks for your work so far on this!  Omitting the systems directories will cause compatibility issues with `rpm5`, which requires that the full tree be declared. `omnibus` [handles this case](https://github.com/chef/omnibus/blob/master/lib/omnibus/packagers/rpm.rb#L273-L287) by writing spec file entries for directories owned by the `filesystem` package that specify the same ownership and permissions they are given in the `filesystem` package.  Is this a change you would have time/inclination to make?

eoly:
Hi @yzl,

After looking a little deeper the issue I was running into in #663 happens because I was doing it wrong. I was trying to install my systemd unit file into /lib/systemd/system but should have been using /usr/lib/systemd/system. See [RHEL Docs](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/System_Administrators_Guide/chap-Managing_Services_with_systemd.html#tabl-Managing_Services_with_systemd-Introduction-Units-Locations)

Manually copying the unit file was working because /lib symlinks to /usr/lib. However the RPM builder code does not handle the symlink properly and results in a conflict when trying to install the package.

I'm not sure that any code changes need to be made. If you agree please feel free to close this out.

Thanks,
Eric

(details below)

```
[root@localhost omnibus-test]# ls -lah /
total 70K
dr-xr-xr-x. 19 root root 4.0K Apr  8 11:29 .
dr-xr-xr-x. 19 root root 4.0K Apr  8 11:29 ..
lrwxrwxrwx.  1 root root    7 Apr  4 17:57 bin -> usr/bin
dr-xr-xr-x.  5 root root 1.0K Apr  8 09:43 boot
drwxr-xr-x. 18 root root 2.9K Apr 13 15:49 dev
drwxr-xr-x. 82 root root 4.0K Apr 13 15:49 etc
drwxr-xr-x.  3 root root 4.0K Apr  4 18:08 home
lrwxrwxrwx.  1 root root    7 Apr  4 17:57 lib -> usr/lib
lrwxrwxrwx.  1 root root    9 Apr  4 17:57 lib64 -> usr/lib64
```

In my project definition I was declaring:

`extra_package_file "/lib/systemd/system/thing.service"` 

Which causes /lib to get added the RPM SPEC like so:

`%dir %attr(0755,root,root) /lib`

That seems to be where the conflict with the filesystem package happens.

By changing the extra_package_file definition of the systemd service to:

`extra_package_file "/usr/lib/systemd/system/thing.service"` 

The RPM SPEC sets the proper permissions on the /usr and /usr/lib dirs:

```
%dir %attr(0755,root,root) /usr
%dir %attr(0555,root,root) /usr/lib
```

The code [here](https://github.com/chef/omnibus/blob/master/lib/omnibus/packagers/rpm.rb#L273-L287) handles the /usr/lib permissions properly but not /lib as it is a symllink.

PR #669, opened 2016-04-27
Add Build Timings to Log Output

PrajaktaPurohit:
Add  build timings to the omnibus logging.

https://github.com/chef/omnibus/pull/669/files?w=1 (Thanks Steven!)

/cc @chef/omnibus-maintainers

### DIFF
--- a/lib/omnibus/logger.rb
+++ b/lib/omnibus/logger.rb
@@ -16,6 +16,9 @@
 
 module Omnibus
   class Logger
+
+    require 'time'
+
     #
     # The amount of padding on the left column.
     #
@@ -135,7 +138,7 @@ module Omnibus
       else
         left = "#{format_severity(severity)} | "
       end
-      "#{left.rjust(LEFT)}#{message}\n"
+      "#{left.rjust(LEFT)}#{Time.now.iso8601()} | #{message}\n"
     end
 
     #
--- a/lib/omnibus/logger.rb
+++ b/lib/omnibus/logger.rb
@@ -24,7 +24,7 @@ module Omnibus
     #
     # @return [Fixnum]
     #
-    LEFT = 40
+    LEFT = 30
 
     #
     # Our custom log levels, in order of severity
--- a/features/commands/manifest.feature
+++ b/features/commands/manifest.feature
@@ -14,7 +14,10 @@ Feature: omnibus manifest
   Scenario: When the project has no software definitions
     When I run `omnibus manifest hamlet`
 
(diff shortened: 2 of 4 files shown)


lib/omnibus/logger.rb
```diff
@@ -135,7 +135,7 @@ def format_message(severity, progname, message)
       else
         left = "#{format_severity(severity)} | "
       end
-      "#{left.rjust(LEFT)}#{message}\n"
+      "#{left.rjust(LEFT)}#{Time.now} | #{message}\n"
```
stevendanna: Looks like changing the output broke a few of our assertions so the tests are failing.  But I'm really glad to have the time in the logs now, it will really help debugging.

stevendanna:
For other reviewers: The hide-whitespace-changes view is useful for this one: https://github.com/chef/omnibus/pull/669/files?w=1

stevendanna:
👍 once the test assertions are adjusted to account for the new output format.

lamont-granquist:
👍

scotthain:
👍 very nice! (pending test adjustment)

lib/omnibus/logger.rb
schisamo: Let's match the formatting Chef uses (via mixlib-log):

```
Time.now.iso8601()
```

Make sure you add a `require 'time'` so the `.iso8601()` method is available.

Example output:

```ruby
irb(main):001:0> require 'time'
=> true
irb(main):002:0> Time.now.iso8601()
=> "2016-04-27T14:31:11-04:00"
```

schisamo:
For those interested, here is some example output from a build executed using this branch of Omnibus:
https://gist.github.com/schisamo/bc0ea2673e02d112cc885682eb7a914b

lib/omnibus/logger.rb
```diff
@@ -21,7 +21,7 @@ class Logger
     #
     # @return [Fixnum]
     #
-    LEFT = 40
+    LEFT = 20
```
schisamo: Maybe we should make this value `30`. As you can see from this test build output the alignment is off for a number of lines:
https://gist.github.com/schisamo/bc0ea2673e02d112cc885682eb7a914b

lib/omnibus/logger.rb
PrajaktaPurohit: 👍

lib/omnibus/logger.rb
PrajaktaPurohit: 👍

PR #670, opened 2016-05-03
Also catch openssl errors when trying to fetch license files so weird openssl errors don't fail our build

tyler-ball:
### Description

We were seeing the following errors failing our ChefDK builds:

```
/opt/languages/ruby/2.1.5/lib/ruby/2.1.0/net/http.rb:920:in `connect': SSL_connect SYSCALL returned=5 errno=0 state=SSLv2/v3 read server hello A (OpenSSL::SSL::SSLError)
```

With this change, we now see the warning:

```
[Licensing] W | 2016-05-03T10:53:04-07:00 | Can not download license file 'https://www.openssl.org/source/license.html' for software 'openssl-customization'.
```

### DIFF
--- a/lib/omnibus/licensing.rb
+++ b/lib/omnibus/licensing.rb
@@ -161,7 +161,8 @@ module Omnibus
                      Errno::ECONNRESET,
                      Errno::ENETUNREACH,
                      Timeout::Error,
-                     OpenURI::HTTPError
+                     OpenURI::HTTPError,
+                     OpenSSL::SSL::SSLError
                 licensing_warning("Can not download license file '#{license_file}' for software '#{name}'.")
               end
             end


danielsdeleo:
👍

tyler-ball:
\cc @tylercloke @danielsdeleo @sersut

sersut:
👍

schisamo:
👍  assuming Travis is happy.

lamont-granquist:
should this do a retry loop -- and maybe just `rescue` (RuntimeError)?

but 👍

ksubrama:
Oh yeah there's an issue where we can't download anything from https://www.openssl.org from within our  omnibus installs on windows.  When writing the openssl software def, I worked around it by manually downloading the openssl binaries, verifying the sha, putting the files in S3 omnibus cache and then using that exclusively everywhere.  I never quite figured out why it was breaking...

danielsdeleo:
@ksubrama As best I can tell our openssl is too old and OpenSSL dot org is requiring a TLS version or cipher suite we don't have.

PR #672, opened 2016-05-04
Actually flush files before performing inode operations

ksubrama:
### Description

A ruby IO.close doesn't flush filesystem buffers.  A file copy and a file write can be considered out-of-order by the underlying file-system.  See also: everyone hating on ext4 when it first came out...

### DIFF
--- a/lib/omnibus/compressors/tgz.rb
+++ b/lib/omnibus/compressors/tgz.rb
@@ -89,6 +89,7 @@ module Omnibus
         while chunk = contents.read(1024)
           tgz.write(chunk)
         end
+        tgz.fsync if tgz.respond_to?(:fsync)
       end
 
       # Copy the .tar.gz into the package directory
--- a/lib/omnibus/download_helpers.rb
+++ b/lib/omnibus/download_helpers.rb
@@ -77,9 +77,14 @@ module Omnibus
         end
 
         file = open(from_url, options)
-        # This is a temporary file. Close and flush it before attempting to copy
-        # it over.
+        # This is a temporary file. You must fsync a file before you can copy it
+        # directly. Just closing/flushing is insufficient as the file-system can
+        # decide to buffer your write and (depending on the ordering-guarantee
+        # flags you have chosen) choose to execute the copy operation before
+        # the contents are written out.
+        file.fsync if file.respond_to?(:fsync)
         file.close
+
         FileUtils.cp(file.path, to_path)
         file.unlink
       rescue SocketError,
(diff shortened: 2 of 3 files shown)


danielsdeleo:
👍

schisamo:
@ksubrama Just a heads up this _could_ throw an exception. I noticed this [tidbit in the docs](http://ruby-doc.org/core-2.2.0/IO.html#method-i-fsync):

> NotImplementedError is raised if the underlying operating system does not support fsync(2).

We need to do a full test build of Chef using this version of Omnibus. We might also want to guard against this exception and fallback to the old behavior in those cases.

ksubrama:
Hmm..  good call.  I think I'll just catch it.

schisamo:
@ksubrama Thx man. We'll also want some unit test coverage for the case that `NotImplementedError ` is thrown, sort of like we do here:
https://github.com/chef/omnibus/blob/acebe9d9a85930422cda6077e93302e9392ef842/spec/unit/package_spec.rb#L62-L68

sethvargo:
This looks really, really sad. Sorry 😦 

👍

lamont-granquist:
Does this solve an actual problem?

I just closed an identical issue in chef/chef as WONTFIX:

 https://github.com/chef/chef/issues/4313

This adds a performance cost (the fsync) while the only problem it solves is correctness on very old versions of ext4 that nobody uses because they're horribly broken.  And the lesson of the ext4 rollout is that if anyone tries that again it will be equally as successful since no matter what POSIX states it breaks the Universe...  We can rely on filesystems not being as clever as that.

So, :-1:

ksubrama:
@lamont-granquist Yeah it (perhaps..) solves a long running issue where we once-in-a-while get checksum mismatches when downloading files from the S3 cache.  The file looks truncated but there's no way the actual download would have been marked complete on a truncated file (the AWS client would have thrown an error).

lamont-granquist:
I would bet significant amounts of money that you've guessed incorrectly.

Are you positive that the AWS client would have thrown an error?  Net::HTTP does not check content-length and will happily write out truncated files, which is why we had to add content-length validation to Chef::HTTP

lamont-granquist:
so the content-length check has been in aws-sdk since 2.0.42:

https://github.com/aws/aws-sdk-ruby/commit/4e165a6ddb08b4a978a369f7b87b80036e0d0558

however, when i manually throw a raise into the aws-sdk-core library in the content-length checking it is never raised by the NetFetcher that downloads the cached omnibus artifacts

lamont-granquist:
yeah you're just using open-uri

lamont-granquist:
I don't have the time to take it over the line but #674 is almost certainly the right fix here.

lamont-granquist:
I'm now willing to be extremely large amounts of money this PR is the wrong fix.

ksubrama:
@lamont-granquist #674 seems like a good start.  Thank you - I'll pick it up when I have the time.

PR #675, opened 2016-05-07
Adds an appx packager to omnibus

mwrock:
This adds a new packager to omnibus that will build appx packages iontended to be installed on Windows Nano Server.


schisamo:
@mwrock re: bullet 2, all Windows packaging concerns on our Jenkins build nodes (and the TK build lab) are installed/configured by `omnibus::_packaging`:
https://github.com/chef-cookbooks/omnibus/blob/master/recipes/_packaging.rb#L64-L69

If we need to add additional SDK tools (or change the ones that are installed) that is currently handled transitively by the `windows-sdk` cookbook.

mwrock:
Totally. Just making a note here so I don't forget to add it there if necessary.

lib/omnibus/signing.rb
```diff
@@ -0,0 +1,157 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+module Omnibus
+  module Signing
```
schisamo: Perhaps we should instead create a `Omnibus::Packager::WindowsBase ` (which extends from `Omnibus::Packager::Base`) class and place these shared methods in it? That would more idiomatically follow how the rest of Omnibus code is laid out. A global module like this would make more sense if we needed to mix in these methods into non-packaging code.

lib/omnibus/exceptions.rb
```diff
@@ -330,9 +330,9 @@ def initialize(version)
     end
   end
 
-  class FailedToTimestampMSI < Error
+  class FailedToTimestampPackage < Error
     def initialize
-      super("Failed to add timestamp to MSI.")
+      super("Failed to add timestamp to Package.")
```
schisamo: Perhaps: `Failed to add timestamp to Windows Package.`

mwrock:
I'm considering rolling the appx packager into the MSI packager. So the MSI packager would build both. On the one hand this would tremendously simplify package building so we dont have to add any logic regarding when to build an appx and when to build an MSI. We just build both all of the time which is fine. The down side would be at package retrieval time. I have not yet investigated if it would be possible/easy to make an omnibus request for JUST an appx or msi if we took this approach. Being able to download just one is a unmovable requirement.

schisamo:
@mwrock Having the Windows build produce both a `*.msi` and `*.appx` is exactly the [approach I recommended taking on FLOW-357](https://chefio.atlassian.net/browse/FLOW-357?focusedCommentId=22608&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-22608). We should be opinionated in this case and follow YAGNI. In this case the thing that won't be needed is producing 1..n packages on any platform except Windows. What we don't want to do is add public APIs and surface area we'll need to support going forward. We'll want to rename `Omnibus::Packager::MSI` to something like `Omnibus::Packager::Windows` to signify it's now about producing more than an MSI.

Also don't worry about how we'll publish or retrieve dual `*.msi` and `*.appx` packages. We (ES) will solve that issue as we roll out support for `*.appx`.

mwrock:
I have replaced the `appx` packaged with a `windows` packager that packages both an MSI and an appx package. *Temporarily* I am leaving the `:msi` packager and keeping windows builders bound to that packager. This should allow us to migrate build projects without breaking builders provided we move forward with this plan:

1. Merge this PR
2. Retofit chef and chef-dk repos (are there others?) adding windows resources
3. Change `Packager` to bind windows to `:windows` packager
4. Remove `:msi` definitions from consuming repos
5. Remove `msi` packager

Thoughts @schisamo ? Am I over thinking this?

mwrock:
@schisamo still a couple details I'm wrapping up (like moving the signing mixin to a base class) but I wanted to get your opinion on ow I'm handling the multi package issue in packager.rb and project.rb. The real awkwardness comes into play with `Omnibus::Compressor::Base`. There is a fair amount of refactoring that would need to be done here to get the compressors working nicely with multiple packages. However right now only windows uses multiple packages and it does not use a compressor. So I cheated and compressor.package just references the first packager in the array. I added an explanatory comment but its a bit gross.

scotthain:
So I'm wondering if we can (instead of what you're doing which makes the project support multiple packages) just make the `Windows` class do 2 things... That way the compressor thing wouldn't be an issue, if we had a compressor for windows, it would have to know about both artifacts that the single `Windows` packager creates. Might simplify things a bit?

I'm late to the party here so if y'all already hashed this out, ignore me.

mwrock:
@scotthain heh. That was iteration #2 on this PR and it ended up proving to be messier that anticipated. The thinking was just as you suggest here, that is would be much simpler but both artifatcs have different staging locations and much more problematic is the `package_name`. Both must have different package names.

I may just go ahead and refactor the compressoers. Its really not that big of an ordeal. We only have a couple. I'm just trying to keep the footprint down here.

scotthain:
Ah ok sorry for the confusion. I wonder if we can refactor the compressor code (to support multiple etc) separately so it's easier to test and review, then integrate it into this? (I'm just spitballing here)

mwrock:
Yeah I do lean toward a separate PR for that.

mwrock:
I've squashed and removed the WIP. Ready for review.

lib/omnibus/packager.rb
```diff
@@ -67,13 +68,13 @@ def for_current_system
         family = "solaris"
       end
       if klass = PLATFORM_PACKAGER_MAP[family]
-        klass
+        [klass].flatten
```
schisamo: Would probably better to not brute force this:

```ruby
klass.is_a?(Array) ? klass : [ klass ]
```

lib/omnibus/packager.rb
```diff
@@ -67,13 +68,13 @@ def for_current_system
...
       else
         log.warn(log_key) do
           "Could not determine packager for `#{family}', defaulting " \
           "to `makeself'!"
         end
-        Makeself
+        [Makeself].flatten
```
schisamo: This `#flatten` seems frivolous.

lib/omnibus/packagers/appx.rb
```diff
@@ -0,0 +1,124 @@
...
+    #     rel builds => 11.14.0.alpha.1 || 11.14.0
+    #
+    # The APPX version spec expects a version that looks like X.Y.Z.W where
+    # X, Y, Z & W are all 32 bit integers.
+    #
+    # @return [String]
+    #
+    def appx_version
+      versions = project.build_version.split(/[.+-]/)
+      "#{versions[0]}.#{versions[1]}.#{versions[2]}.#{project.build_iteration}"
```
schisamo: This is a little more readable:

```ruby
major, minor, patch = project.build_version.split(/[.+-]/)
[major, minor, patch, project.build_iteration].join('.')
```

schisamo:
@mwrock Left a few comments for you....overall things look great. 

We'll need to ensure we update the Omnibus project generator: 

* add CLI option for generating APPX package assets - https://github.com/chef/omnibus/blob/master/lib/omnibus/generator.rb#L53-L56
* add a `create_appx_assets` method - https://github.com/chef/omnibus/blob/master/lib/omnibus/generator.rb#L127-L140

lib/omnibus/packager.rb
mwrock: this does the same thing as `klass.is_a?(Array) ? klass : [ klass ]`. it forces the value to an array. and in case it just made a "double array" it flattens it back to a single array.

lib/omnibus/packagers/appx.rb
mwrock: cool. I basically did a direct copy/paste here from the `msi_version` which reminds me since we made a base class, thats the propper home for this and I'll change the implementation to match above

lib/omnibus/packager.rb
schisamo: But `Makeself` is a class and will never be an Array so the `#flatten` in this case will never be needed.

mwrock: oh i see what you are saying. yes. frivilous indeed

mwrock:
ok @schisamo I made the recommended changes including the generator stuff

lib/omnibus/packagers/appx.rb
```diff
@@ -0,0 +1,104 @@
...
+    #
+    def write_manifest_file
+      render_template(resource_path('AppxManifest.xml.erb'),
+        destination: "#{windows_safe_path(project.install_dir)}/AppxManifest.xml",
+        variables: {
+          name:            project.package_name,
+          friendly_name:   project.friendly_name,
+          version:         windows_package_version,
+          maintainer:      project.maintainer,
+          certificate_subject: certificate_subject,
```
schisamo: The `certificate_subject` should only be set if we are signing the package correct?

mwrock: we still want to call that method so that the variable gets assigned to something in the template or else the appx generation will bomb. It will default to the project name if there is no cert.

lib/omnibus/packagers/appx.rb
```diff
@@ -0,0 +1,104 @@
...
+    def pack_command(appx_file)
+      "makeappx.exe pack /d \"#{windows_safe_path(project.install_dir)}\" /p #{appx_file}"
+    end
+
+    #
+    # Get the certificate subject of the signing identity
+    #
+    # @return [String]
+    #
+    def certificate_subject
```
schisamo: Should this method be part of `Omnibus::Packager::WindowsBase` with the rest of the signing methods?

mwrock: yeah I'll move it.

lib/omnibus/packagers/appx.rb
schisamo: Sounds good.

mwrock:
BTW: appveyor failure is unrelated to this PR `http://curl.haxx.se/ca/cacert.pem` is down.

schisamo:
@mwrock A few final items:

* We shouldn't need `resources/appx/CustomActionFastMsi.CA.dll` correct? That DLL should only be used by the MSI/WiX stuff in `Omnibus::Packager::MSI`.
* The `resources/appx/assets/chef.png` should be generic maybe `omnibus.png` (we can use the Chef Software Inc version, `chef.png`, in [chef/chef](https://github.com/chef/chef) and [chef/chef-dk](https://github.com/chef/chef-dk)). Take a look at the assets we have in the `resources/msi/assets` folder:
https://github.com/chef/omnibus/tree/master/resources/msi/assets
* We should introduce the new `resources/appx/*` stuff in [chef/chef](https://github.com/chef/chef) (`omnibus/resources/chef/appx`) and [chef/chef-dk](https://github.com/chef/chef-dk) (`omnibus/resources/chefdk/appx`) before we merge this branch as both projects auto-update to the latest version of Omnibus.

mwrock:
@schisamo 

> We shouldn't need resources/appx/CustomActionFastMsi.CA.dll correct? That DLL should only be used by the MSI/WiX stuff in Omnibus::Packager::MSI.
That is correct.

Yeah the icon is super dumb and can really be anything because no one will ever see it. Its mainly used for windows store apps and has no relevance to windows server apps, but the schema requires it. Making it as generic as possible is good. I'll swap the image.

> We should introduce the new resources/appx/* stuff in chef/chef (omnibus/resources/chef/appx) and chef/chef-dk (omnibus/resources/chefdk/appx) before we merge this branch as both projects auto-update to the latest version of Omnibus.
Wouldn't chef and chefdk "inherit" the resources from the omnibus resources? They will be exactly the same.

scotthain:
@mwrock I think (and I don't want to speak for @schisamo) it should be a generic logo as in 'not Chef Software Inc' logo, since non-chef projects use omnibus as a framework. That would explain Seth's comment about putting them in chef/chef and chef/chefdk projects, respectively. I'm sure Seth will clarify.

mwrock:
oh gotcha @scotthain. So I'm thinking we change the icon to a bus and then we really dont need anything extra in the chef or chefdk repos. They can inherit the artifacts from omnibus and the project template variable replacements will provide all the customization we need.

mwrock:
To be clear. an image, any image just has to be in the package but they are never rendered.

scotthain:
@mwrock how about a "clear" png? :)

mwrock:
actually thats a great idea.

mwrock:
clear.gif it is.

schisamo:
@mwrock Per my earlier comment we still need `resources/appx/CustomActionFastMsi.CA.dll` removed.

mwrock:
argh its gone now @schisamo . sublime was hiding it from me.

PrajaktaPurohit:
awesome! +1

PR #676, opened 2016-05-13
strip trailing whitespace from certificate subject

mwrock:
Ugh. Just noticed that #675 has an issue with grabbing the certificate subject that did not manifest in my testing before. It appears the powershell returns the string with trailing whitespace which then kills `makeappx.exe`. This strips the whitesspace and adds a test for that.


scotthain:
👍

schisamo:
👍

PR #677, opened 2016-05-16
do not package appx on OS versions less than 2012

mwrock:
Windows server versions less than 2012 do not support the running of `MakeAPPX.exe`. This removes APPX from the available packagers on those versions.

### DIFF
--- a/lib/omnibus/packager.rb
+++ b/lib/omnibus/packager.rb
@@ -68,7 +68,15 @@ module Omnibus
         family = "solaris"
       end
       if klass = PLATFORM_PACKAGER_MAP[family]
-        klass.is_a?(Array) ? klass : [ klass ]
+        package_types = klass.is_a?(Array) ? klass : [ klass ]
+
+        if package_types.include?(APPX) &&
+          !Chef::Sugar::Constraints::Version.new(version).satisfies?('>= 6.2')
+          log.warn(log_key) { "APPX generation is only supported on Windows versions 2012 and above" }
+          package_types = package_types - [APPX]
+        end
+
+        package_types
       else
         log.warn(log_key) do
           "Could not determine packager for `#{family}', defaulting " \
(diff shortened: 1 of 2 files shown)


lib/omnibus/packager.rb
```diff
@@ -68,7 +68,15 @@ def for_current_system
         family = "solaris"
       end
       if klass = PLATFORM_PACKAGER_MAP[family]
-        klass.is_a?(Array) ? klass : [ klass ]
+        package_types = klass.is_a?(Array) ? klass : [ klass ]
+
+        if package_types.include?(APPX) &&
+          !Chef::Sugar::Constraints::Version.new(version).satisfies?('>= 6.2')
```
schisamo: I believe the following less-verbose check should work also:

```ruby
if package_types.include?(APPX) && platform_version.satisfies?('>= 6.2')
```

`platform_version` is part of Chef Sugar's DSL and is brought in by way of the `Omnibus::Sugarable` mixin:
https://github.com/chef/omnibus/blob/master/lib/omnibus/sugarable.rb#L29-L36
https://github.com/sethvargo/chef-sugar/blob/master/lib/chef/sugar/platform.rb#L276-L286

mwrock: yeah I had wondered the same but got `undefined local variable or method `platform_version' for Omnibus::Packager:Module` and also saw the solaris checks were using the longer form. I'll look to see why `platform_version` does not take. Maybe something to do with how the [packager module is pulled in.

mwrock: While I investigate, I dont think this should block especially since this is blocking green builds

schisamo: @mwrock Totally agree but we need to get Travis and Appveyor green before merging.

mwrock: appveyor is network stuff unrelated to this. Working on travis now.

mwrock:
travis is green @schisamo

PR #678, opened 2016-05-16
Add :fatal_licensing_warnings configuration option

sersut:
### Description

Adding a new configuration option, as previously discussed in https://github.com/chef/omnibus/pull/662, that fails the build when we encounter a licensing warning. 

I have changed one of the warnings since practically any project who uses `omnibus-software` will hit this that warning due to this line:

https://github.com/chef/omnibus-software/blob/master/config/software/config_guess.rb#L23

Our goal is to introduce some form of configurable white-listing or black-listing in the future and convert this to warning again. 

You can see an example output both with this option set and not set below (also full logs - with redacted build stages - [here](https://gist.github.com/sersut/c161c35d6f6152a69002812e8bc498c0).)

```
$ bundle exec omnibus build harmony
...
              [Licensing] W | 2016-05-16T12:56:01-07:00 | Project 'harmony' does not contain licensing information.
 [Software: config_guess] W | 2016-05-16T12:56:01-07:00 | Version master for software config_guess was not parseable. Comparison methods such as #satisfies? will not be available for this version.
              [Licensing] W | 2016-05-16T12:56:01-07:00 | Software 'discord' does not contain licensing information.
              [Licensing] I | 2016-05-16T12:56:01-07:00 | Software 'config_guess' uses license 'GPL-3.0 (with exception)' which is not one of the standard licenses identified in https://opensource.org/licenses/alphabetical. Consider using one of the standard licenses.
              [Licensing] W | 2016-05-16T12:56:01-07:00 | Software 'popt' does not contain licensing information.
              [Licensing] W | 2016-05-16T12:56:01-07:00 | Software 'rsync' does not contain licensing information.
...

$ bundle exec omnibus build harmony --override fatal_licensing_warnings:true
...
              [Licensing] W | 2016-05-17T10:54:10-07:00 | Software 'discord' does not contain licensing information.
              [Licensing] I | 2016-05-17T10:54:10-07:00 | Software 'config_guess' uses license 'GPL-3.0 (with exception)' which is not one of the standard licenses identified in https://opensource.org/licenses/alphabetical. Consider using one of the standard licenses.
              [Licensing] W | 2016-05-17T10:54:10-07:00 | Software 'popt' does not contain licensing information.
              [Licensing] W | 2016-05-17T10:54:10-07:00 | Software 'rsync' does not contain licensing information.
Encountered error(s) with project's licensing information.
Failing the build because :fatal_licensing_warnings is set in the configuration.
Error(s):

    Project 'harmony' does not contain licensing information.
    Software 'discord' does not contain licensing information.
    Software 'popt' does not contain licensing information.
    Software 'rsync' does not contain licensing information.
...
```
/cc @chef/omnibus-maintainers

### DIFF
--- a/lib/omnibus/config.rb
+++ b/lib/omnibus/config.rb
@@ -446,7 +446,7 @@ module Omnibus
     #
     # For example:
     #
-    #     /PATH/files/my_map_file 
+    #     /PATH/files/my_map_file
     #
     # @return [String, nil]
     default(:solaris_linker_mapfile, "files/mapfiles/solaris")
@@ -508,6 +508,11 @@ module Omnibus
         3
       end
     end
+
+    # Fail the build or warn when build encounters a licensing warning.
+    #
+    # @return [true, false]
+    default(:fatal_licensing_warnings, false)
 
     # --------------------------------------------------
     # @!endgroup
--- a/lib/omnibus/exceptions.rb
+++ b/lib/omnibus/exceptions.rb
@@ -335,4 +335,20 @@ EOF
       super("Failed to sign Windows Package.")
     end
   end
+
+  class LicensingError < Error
+    def initialize(errors)
+      @errors = errors
+    end
+
+    def to_s
+      <<-EOH
+Encountered error(s) with project's licensing information.
+Failing the build because :fatal_licensing_warnings is set in the configuration.
+Error(s):
+
+    #{@errors.join("\n    ")}
+EOH
+    end
+  end
 end
--- a/lib/omnibus/licensing.rb
+++ b/lib/omnibus/licensing.rb
@@ -41,11 +41,19 @@ module Omnibus
     attr_reader :project
 
     #
+    # The warnings encountered while preparing the licensing information
+    #
+    # @return [Array<String>]
+    #
+    attr_reader :licensing_warnings
+
+    #
     # @param [Project] project
     #   the project to create licenses for.
     #
     def initialize(project)
       @project = project
+      @licensing_warnings = []
     end
 
     #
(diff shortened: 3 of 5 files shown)


lib/omnibus/project.rb
```diff
@@ -1073,16 +1073,16 @@ def build
...
+      # end
 
       write_json_manifest
       write_text_manifest
       Licensing.create!(self)
       HealthCheck.run!(self)
-      package_me
-      compress_me
+      # package_me
+      # compress_me
```
schisamo: Looks like some commented out code was inadvertently commited.

sersut: Argh you got me :)

scotthain:
👍

schisamo:
@sersut Would it be possible to "collect" all failures and throw a single exception? I think the UX would be much nicer as I would see a list of all my licensing violations in the exception's message. 

It would also be nice to eventually expose an `omnibus lint` subcommand that allows us to run things like the licensing check in the _verify_ stage of a pipeline.

lamont-granquist:
config_guess is a build-dependency that does include its source code in anything included in the artifact.

there should really be a way to simply omit licensing information for it.  it is irrelevant to the artifact that we ship.

i'm thinking the software config should look a bit more like:

```ruby
license "GPL-3.0"
# this is a build-time dep whose licensing does not affect the shipped artifact
omit_licensing_information true
```

then it could be an error not to have a valid license statement when the project is parsed, but any other use of the license information would get skipped.

sersut:
Agreed @schisamo. Updated the code to raise one error for everything at the end. 

@lamont-granquist even if we have that functionality we probably will have a software component that does not adhere to the standard set of licenses. The issue is not about the build time / run time components, it is about the fact that we do not have granular control over what is permitted and what is not (which we are hoping to address in the future).

schisamo:
👍

yzl:
👍

PR #680, opened 2016-05-20
Jam/refactor packager

jmink:
### Description

This is a refactor of the packager.rb class to make it easier to add OS specific logic.  Functionality should be unchanged.

/cc @chef/omnibus-maintainers


schisamo:
@jmink Could you give some background on why this change is needed?

lib/omnibus/packager.rb
```diff
@@ -30,24 +28,131 @@ module Packager
...
+
+        [Makeself]
+      end
+    end
+
+    class Solaris2Platform < Platform
+      require 'omnibus/packagers/solaris'
+      require 'omnibus/packagers/ips'
+
+      def supported_packagers
```
seth: What happens if we encounter an unsupported version less than 5.10? I'm assuming we end up failing to build because we don't have a packager and so it would be better to have that failure happen here.

Style-wise, I'd prefer an if-as-expression style over the multiple `return`s.

lib/omnibus/packager.rb
```diff
@@ -59,31 +164,7 @@ module Packager
...
-
-        package_types
-      else
-        log.warn(log_key) do
-          "Could not determine packager for `#{family}', defaulting " \
-          "to `makeself'!"
-        end
-        [Makeself]
-      end
+      return Platform.create(Ohai).supported_packagers
```
seth: The `return` isn't needed.

spec/unit/packager_spec.rb
```diff
@@ -1,70 +1,147 @@
...
+    describe '.supported_packagers' do
+      it 'returns Makeself' do
+        platform_info = {'platform_family' => 'nonexistant', 'platform_version' => '0.1.1'}
+        instance = Packager::DefaultPlatform.new(platform_info)
+        expect(instance.supported_packagers).to eq([Packager::Makeself])
+      end
+    end
+  end
+
+  describe Packager::Solaris2Platform do
```
seth: maybe a test for unsupported version here?

lib/omnibus/packager.rb
```diff
@@ -30,24 +28,131 @@ module Packager
...
+
+        begin
+          name = Packager.const_get(class_name(platform_name))
+        rescue NameError
+          name = DefaultPlatform
+        end
+        name.new(platform_information)
+      end
+
+      # This function should be overritten by its child class
```
sethvargo: There's a special YARD syntax for specifying this is an abstract method/class.

lib/omnibus/packager.rb
jmink: It used to be a key look up error.  I'll change it to raising a meaningful exception.

jmink: Actually running some tests it returns Makeself.  I've updated the code to do that and switched to a case statement instead of short circuiting with returns.

jmink: Fixed (Though testing shows that the old code returned Makeself, not an exception, so that you for pointing that out)

PR #681, opened 2016-05-20
Add ChefStyle!

schisamo:
There's a new cop in town! This brings Omnibus in line with Chef Software Inc's other projects that are using ChefStyle for linting. You can read more about ChefStyle here: https://github.com/chef/chefstyle

/cc @chef/omnibus-maintainers


sersut:
Yes! 👍

scotthain:
cop roooock

lamont-granquist:
so something you should be aware of is that chefstyle has a ton of cops (like appx 100 of them, which is a little under half i think) turned off -- not by policy but because the work hasn't been done to fix the chef/chef codebase and turn them on.

i created my own lcgstyle gem for my own use which is chefstyle with all those cops turned on.

we probably want to consider having a flag to chefstyle to make it operate in that mode (or alternatively make that the default and make a flag specifically for chef/chef)

or i suppose alternatively we could just have a massive .rubocop.yml TODO in chef/chef that disables everything we haven't done yet...

schisamo:
@lamont-granquist Considering we had no linting enabled in this repo previously I think we are OK starting out with a minimal set of Cops enabled. I think its more important we apply the same set of rules to all of our Ruby repos, even if that set of rules is small.

yzl:
👍

lamont-granquist:
ah, well then proceed...  =)

👍

PR #687, opened 2016-06-16
convert entry name symbols to strings for repo_path join

srenatus:
### Description

The introduction of symbolized keys in the metadata diffs broke change log generation:

```interactive
$ b/omnibus changelog generate --starting-manifest=../opt/opscode/version-manifest.json --ending-manifest=../version-manifest.json
                    [CLI] I | 2016-06-16T11:38:59+02:00 | Using config from 'omnibus.rb'
     [Command::ChangeLog] I | 2016-06-16T11:38:59+02:00 | Using config from 'omnibus.rb'
## 12.7.0 (2016-06-16)
### Components
New Components
* veil-gem (master)

Updated Components
* config_guess (706fbe57 -> ddd7f330)

/Users/stephan/Src/chef/omnibus/lib/omnibus/changelog_printer.rb:53:in `join': no implicit conversion of Symbol into String (TypeError)
        from /Users/stephan/Src/chef/omnibus/lib/omnibus/changelog_printer.rb:53:in `block in print_updated_components'
        from /Users/stephan/Src/chef/omnibus/lib/omnibus/changelog_printer.rb:49:in `each'
        from /Users/stephan/Src/chef/omnibus/lib/omnibus/changelog_printer.rb:49:in `print_updated_components'
        from /Users/stephan/Src/chef/omnibus/lib/omnibus/changelog_printer.rb:33:in `print_components'
        from /Users/stephan/Src/chef/omnibus/lib/omnibus/changelog_printer.rb:14:in `print'
        from /Users/stephan/Src/chef/omnibus/lib/omnibus/cli/changelog.rb:82:in `generate'
```

This fixes it. :)

### DIFF
--- a/lib/omnibus/changelog_printer.rb
+++ b/lib/omnibus/changelog_printer.rb
@@ -49,8 +49,8 @@ module Omnibus
       diff.updated.each do |entry|
         puts sprintf("* %s (%.8s -> %.8s)",
                      entry[:name], entry[:old_version], entry[:new_version])
-        repo_path = ::File.join(source_path, entry[:name])
-        if entry[:source_type] == "git" && ::File.directory?("#{repo_path}/.git")
+        repo_path = ::File.join(source_path, entry[:name].to_s)
+        if entry[:source_type] == :git && ::File.directory?("#{repo_path}/.git")
           cl = ChangeLog.new(entry[:old_version], entry[:new_version], GitRepository.new("#{repo_path}"))
           print_changelog(cl, 2)
         end
--- a/spec/unit/changelogprinter_spec.rb
+++ b/spec/unit/changelogprinter_spec.rb
@@ -1,0 +1,130 @@
+require "spec_helper"
+require "omnibus/changelog_printer"
+require "omnibus/manifest_diff"
+
+module Omnibus
+  describe ChangeLogPrinter do
+    describe "#print" do
+      def manifest_entry_for(name, dv, lv, source_type = :local)
+        Omnibus::ManifestEntry.new(name, { :described_version => dv,
+                                           :locked_version => lv,
+                                           :locked_source => {
+                                             :git => "git://#{name}@example.com" },
+                                           :source_type => source_type,
+        })
+      end
+
+      let(:changelog) { double(ChangeLog,
+                               changelog_entries: %w{entry1 entry2},
+                               authors: %w{alice bob}) }
+      let(:git_changelog) { double(ChangeLog,
+                                   changelog_entries:
+                                   %w{sub-entry1 sub-entry2}) }
+      let(:now) { double(Time) }
+      let(:emptydiff) { EmptyManifestDiff.new }
+      let(:old_manifest) {
+        m = Manifest.new()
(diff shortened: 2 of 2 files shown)


stevendanna:
👍

schisamo:
@srenatus Is there any test coverage we could add with this fix?

sethvargo:
Fix looks good, but I agree with @schisamo - would like to see a test to prevent a regression.

srenatus:
@sethvargo, @schisamo: I've backfilled some test for ChangeLogPrinter, but these don't actually cover the bug fixed there. To achieve this, there should be _more-than-unit_-test combining [this](https://github.com/chef/omnibus/blob/master/spec/unit/manifest_diff_spec.rb#L14-L30) with the output matching for [new, updated, removed components](https://github.com/chef/omnibus/blob/e48c7699ca3fbd1362cddce12e040c84c85c420c/spec/unit/changelogprinter_spec.rb#L75-L126).  Not sure where the best place to put this would be, though. `spec/functional/`?

Also no clue why appveyor fails. :/

schisamo:
@srenatus I'll try re-running the AppVeyor job. I also think `spec/functional` sounds like a reasonable place for your _more-than-unit_ test.

lamont-granquist:
We actually have lots of more-than-unit tests in unit tests.

If the test just requires collaboration between tightly coupled objects so that there's more than one object in the system-under-test (SUT) then its still a unit test.

Generally functional tests are more end-to-end for a given bit of functionality that doesn't interact with external things (i.e. no integration between the client and a chef-zero server).  The func tests are often restricted so that they can only run on a given operating system (e.g. when func testing dpkg/apt/yum/etc by really running those apps).

As long as the System-Under-Test involves a few tightly coupled ruby objects, along with possibly the filesystem, and is not OS-specific then its probably a unit test.  We have some windows and unix specific tests in unit tests as well when it just becomes overly fussy to stub out the unix-ness or windows-ness of the tests.

srenatus:
@lamont-granquist @schisamo: thanks, I'll change those tests to use two manifests instead of the mocked manifest diff then (on Monday). _This_ will also cover the issue fixed here.

srenatus:
Updated tests as discussed, [found another bug](https://github.com/chef/omnibus/pull/687/commits/4bd05bca21408d1200b993d094fb1a3914e57ecb#diff-479313a2e756a8670283368b34156f9aR53) in the process. 🎉

srenatus:
Appveyor 👎 

```
    Errno::ETIMEDOUT:
       A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond. - connect(2) for "pilotfiber.dl.sourceforge.net" port 80
     # ./lib/omnibus/core_extensions/open_uri.rb:51:in `open_uri'
```

ksubrama:
The appveyor test can be flaky - it tries to access external sites and can time out.  I just kicked it off again.

srenatus:
@ksubrama thanks for kicking appveyor; it seems like it wasn't hard enough :(

@ksubrama @lamont-granquist @schisamo LGTM anyways?

marcparadise:
:+1:

PR #695, opened 2016-07-26
More msys2 upgrade fallout.

ksubrama:
- Be very conservative when putting items on the path when invoking
  commands that involve embedded ruby.
- Be accurate when logging environment variables that are to be used
  within msys2 shells.
- Handle new flags in the mingw cookbook that let you pass env variable
  prefixes into msys2 shells.

---

### DIFF
--- a/lib/omnibus/builder.rb
+++ b/lib/omnibus/builder.rb
@@ -81,6 +81,10 @@ module Omnibus
       warn_for_shell_commands(command)
 
       build_commands << BuildCommand.new("Execute: `#{command}'") do
+        # If we expect the command to run cleanly in windows within a limited
+        # msys shell, we also expect it to run cleanly on other platforms
+        # without the assistance of our chef or ruby binaries.
+        options[:clean_ruby_path] = true if options[:in_msys_bash]
         shellout!(command, options)
       end
     end
@@ -245,6 +249,7 @@ module Omnibus
 
       patches << patch_path
       options[:in_msys_bash] = true
+      options[:clean_ruby_path] = true
       build_commands << BuildCommand.new("Apply patch `#{source}'") do
         shellout!(patch_cmd, options)
       end
@@ -298,6 +303,7 @@ module Omnibus
     def ruby(command, options = {})
       build_commands << BuildCommand.new("ruby `#{command}'") do
         bin = embedded_bin("ruby")
+        options[:clean_ruby_path] = true
         shellout!("#{bin} #{command}", options)
       end
     end
@@ -315,6 +321,7 @@ module Omnibus
     def gem(command, options = {})
       build_commands << BuildCommand.new("gem `#{command}'") do
         bin = embedded_bin("gem")
+        options[:clean_ruby_path] = true
         shellout!("#{bin} #{command}", options)
       end
     end
@@ -335,6 +342,7 @@ module Omnibus
     def bundle(command, options = {})
       build_commands << BuildCommand.new("bundle `#{command}'") do
         bin = embedded_bin("bundle")
+        options[:clean_ruby_path] = true
         shellout!("#{bin} #{command}", options)
       end
     end
@@ -368,6 +376,7 @@ module Omnibus
         # Ensure the main bin dir exists
         FileUtils.mkdir_p(bin_dir)
 
+        options[:clean_ruby_path] = true
         shellout!("#{appbundler_bin} '#{app_software.project_dir}' '#{bin_dir}'", options)
       end
     end
@@ -386,6 +395,7 @@ module Omnibus
     def rake(command, options = {})
       build_commands << BuildCommand.new("rake `#{command}'") do
         bin = embedded_bin("rake")
+        options[:clean_ruby_path] = true
         shellout!("#{bin} #{command}", options)
       end
     end
@@ -795,14 +805,6 @@ module Omnibus
       # Make sure the PWD is set to the correct directory
       # Also make a clone of options so that we can mangle it safely below.
       options = { cwd: software.project_dir }.merge(options)
-
-      if options.delete(:in_msys_bash) && windows?
-        # Mixlib will handle escaping characters for cmd but our command might
-        # contain '. For now, assume that won't happen because I don't know
-        # whether this command is going to be played via cmd or through
-        # ProcessCreate.
-        command_string = "bash -c \'#{command_string}\'"
-      end
 
       # Set the log level to :info so users will see build commands
       options[:log_level] ||= :info
--- a/lib/omnibus/builder.rb
+++ b/lib/omnibus/builder.rb
@@ -281,6 +281,14 @@ module Omnibus
     expose :windows_safe_path
 
     #
+    # (see Util#to_msys2_path)
+    #
+    def to_msys2_path(*pieces)
+      super
+    end
+    expose :to_msys2_path
+
+    #
     # @!endgroup
     # --------------------------------------------------
 
--- a/lib/omnibus/software.rb
+++ b/lib/omnibus/software.rb
@@ -674,7 +674,7 @@ module Omnibus
           arch_flag = windows_arch_i386? ? "-m32" : "-m64"
           opt_flag = windows_arch_i386? ? "-march=i686" : "-march=x86-64"
           {
-            "LDFLAGS" => "-L#{install_dir}/embedded/lib #{arch_flag} -fno-lto",
+            "LDFLAGS" => "-L#{install_dir}/embedded/lib #{arch_flag} -Wl,-rpath,#{install_dir}/embedded/lib",
             # We do not wish to enable SSE even though we target i686 because
             # of a stack alignment issue with some libraries. We have not
             # exactly ascertained the cause but some compiled library/binary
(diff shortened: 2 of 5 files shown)


ksubrama:
@mwrock @chef/engineering-services

ksubrama:
I still need to fix the current tests (and possibly write one for util.rb now that that function is a bit fat and has some functionality.

lib/omnibus/util.rb
```diff
@@ -220,5 +290,24 @@ def create_link(a, b)
...
+    def filter_paths
+      @filter_paths ||=
+        begin
+          # Any alternate binaries or gems from the users environment.
+          filters = Gem.paths.path.dup
+          # Any possible embedded ruby in chef products.
+          if windows?
+            filters << "C:/opscode"
+          else
+            filters << "/opt/chef" << "/opt/chefdk" << "/opt/delivery-cli"
```
schisamo: There should be no Chef Software Inc specifics in this code base.

lib/omnibus/util.rb
```diff
@@ -65,18 +99,54 @@ def shellout(*args)
...
+      path_keys = options[:environment].keys.grep(/\Apath\Z/i)
+      if path_keys.length > 1
+        raise InvalidValue.new("shellout", "receive environment without duplicate PATH values")
+      elsif path_keys.length == 1 && path_keys.first != path_key
+        options[:environment][path_key] = options[:environment].delete(path_keys.first)
+      end
+
+      # Try our best to get rid of the "outer" omnibus ruby and any current
+      # chef-client/chef-dk installations from the path.
+      if options.delete(:clean_ruby_path)
```
schisamo: Why is this needed? We already do clean the outer Ruby environment:
https://github.com/chef/omnibus/blob/29d390ec93709ceec2667038aed6769ee29a0646/lib/omnibus/builder.rb#L833
https://github.com/chef/omnibus/blob/29d390ec93709ceec2667038aed6769ee29a0646/lib/omnibus/builder.rb#L877-L911

ksubrama: That cleans things that are bundler related.  It doesn't take out C:\languages\ruby\... from the path which is a completely incompatible devkit sitting in the path.  It also doesn't remove C:\opscode\angrychef\bin which contains a libz, libarchive, liblzma among other things.  In the course of development, I found doing this helpful to completely eliminate a certain class of problems that I needed to investigate.  Now, if I get a linking related issue and I suspect that it's due to a mismatch, it can only be because I'm accidentally picking up an msys2 binary (and I can't really take those out of the path for obvious reasons).

lib/omnibus/util.rb
ksubrama: Hmm - I could make this a flag in config.rb and configure it in chef/chef-dk.  It does feel like duplicated code though.  Could I reference the chef products in the flag defaults?

schisamo: Can't you just use [`Omnibus::Project#install_dir`](https://github.com/chef/omnibus/blob/184c3f7916f0deed7f1b25ecc755c332729fc2a8/lib/omnibus/project.rb#L157-L185) as the thing you filter out? If you need to filter anything out any install_dir's for other Omnibus projects then it means the `PATH` being used to execute the Omnibus build is too promiscuous.

ksubrama: No no, I want #install_dir.  I want to remove all the other rubies and devkits that potentially end up on the path.

schisamo: That's fine but we should not be playing whack-a-mole with every bad thing that could end up in the `PATH`. That is a losing battle which we'll never cover every case for. Perhaps you are better off building up a fresh/clean `PATH` for the affected functions/commands from `Omnibus::Builder`. It would be preferable to add `C:\msys2\bin`, `C:\msys2\mingw32\bin`, and `C:\msys2\usr\bin` to a clean `PATH` (assuming they exist). We should also isolate this special behavior to Windows only...no reason to muck with *nix because of Windows/MSYS2 sins.

lib/omnibus/software.rb
```diff
@@ -653,7 +654,7 @@ def with_standard_compiler_flags(env = {}, opts = {})
           arch_flag = windows_arch_i386? ? "-m32" : "-m64"
           opt_flag = windows_arch_i386? ? "-march=i686" : "-march=x86-64"
           {
-            "LDFLAGS" => "-L#{install_dir}/embedded/lib #{arch_flag} -fno-lto",
+            "LDFLAGS" => "-L#{install_dir}/embedded/lib #{arch_flag} -Wl,-rpath,#{install_dir}/embedded/lib",
```
ksubrama: CRIT

Unsure if this helps...  it probably doesn't but this is a codechange that did get me past that build issues.  Don't know if it just changed timing of race conditions....

lib/omnibus/software.rb
```diff
@@ -667,9 +668,7 @@ def with_standard_compiler_flags(env = {}, opts = {})
             # TODO: This was true of our old TDM gcc 4.7 compilers. Is it still
             # true with mingw-w64?
             #
-            # XXX: Temporarily turning -O3 into -O2 -fno-lto to work around some
-            # weird linker issues.
-            "CFLAGS" => "-I#{install_dir}/embedded/include #{arch_flag} -O2 -fno-lto #{opt_flag}",
+            "CFLAGS" => "-I#{install_dir}/embedded/include #{arch_flag} -O3 #{opt_flag}",
```
ksubrama: Good to have:  the original fuckery was speculative anyway.  This just puts us back into O3.

lib/omnibus/software.rb
```diff
@@ -708,12 +707,23 @@ def with_standard_compiler_flags(env = {}, opts = {})
         extra_linker_flags["LD_OPTIONS"] = ld_options
       end
 
+      # Always want to favor pkg-config from embedded location to not hose
+      # configure scripts which try to be too clever and ignore our explicit
+      # CFLAGS and LDFLAGS in favor of pkg-config info.
+      pkg_config_flags = {
```
ksubrama: CRIT

lib/omnibus/software.rb
```diff
@@ -708,12 +707,23 @@ def with_standard_compiler_flags(env = {}, opts = {})
...
+        "PKG_CONFIG_PATH" => "#{install_dir}/embedded/lib/pkgconfig",
+      }
+
+      # On windows, make this an msys style path because autotools don't grok
+      # windows paths and use : as a path separator. See the mingw cookbook for
+      # how the flag gets passed into the msys2 shell.
+      if windows?
```
ksubrama: CRIT

lib/omnibus/software.rb
```diff
@@ -732,7 +742,14 @@ def with_standard_compiler_flags(env = {}, opts = {})
     def with_embedded_path(env = {})
       paths = ["#{install_dir}/bin", "#{install_dir}/embedded/bin"]
       path_value = prepend_path(paths)
-      env.merge(path_key => path_value)
+      new_env = env.merge(path_key => path_value)
+
+      # Extra path elements for msys2.
+      if windows?
```
ksubrama: CRIT

ksubrama:
@jkeiser I marked things that are "critical" with CRIT above.  Those change are broken in windows with a new msys as of a week ago.

@schisamo @scotthain  John was looking into extracting a minimal "we are absolutely borked if we don't have this" subset of my changes.  It should hopefully be much much smaller and we can either merge those to master or pin to that branch instead of this one with the functional tests failing.

lib/omnibus/builder.rb
```diff
@@ -276,6 +281,14 @@ def windows_safe_path(*pieces)
     expose :windows_safe_path
 
     #
+    # (see Util#to_msys2_path)
+    #
+    def to_msys2_path(*pieces)
```
ksubrama: CRIT: Needed by liblzma fix.

lib/omnibus/util.rb
```diff
@@ -122,6 +194,18 @@ def windows_safe_path(*pieces)
     end
 
     #
+    # Takes a ruby style path (e.g. C:/foo/bar) and coverts it to an msys path (/C/foo/bar).
+    #
+    # @param [String, Array<String>] pieces
+    #   the pieces of the path to join and fix
+    # @return [String]
+    #
+    def to_msys2_path(*pieces)
```
ksubrama: CRIT: needed by liblzma fix.

stevendanna:
@chef/omnibus-maintainers A number of our projects are now running directly from this branch AND from a branch of omnibus-software.  Is getting this merged inflight? Is there help that can be offered to those working on it?

tduffield:
Rebased on master

mwrock:
It is definitely at the tip of our backlog to address. There is some untangling of the tool chains we need to undertake as part of that work. We definitely acknowledge this as a high priority.

rhass:
This PR can be closed once #759 has been merged, however the branch will need to remain until all projects have been properly migrated over to build from master.

rhass:
I think we can kill this PR now, yes?

ksubrama:
Oh god yes!

PR #699, opened 2016-08-03
Correct p5p package name

scotthain:
### Description

@chef/engineering-services

### DIFF
--- a/lib/omnibus/packagers/ips.rb
+++ b/lib/omnibus/packagers/ips.rb
@@ -81,7 +81,8 @@ module Omnibus
     # @see Base#package_name
     #
     def package_name
-      "#{safe_base_package_name}.p5p"
+      version = project.build_version.split(/[^\d]/)[0..2].join(".")
+      "#{safe_base_package_name}-#{version}-#{project.build_iteration}.#{safe_architecture}.p5p"
     end
 
     #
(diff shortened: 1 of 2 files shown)


yzl:
👍 pending tests

lib/omnibus/packagers/ips.rb
```diff
@@ -81,7 +81,16 @@ def publisher_prefix(val = NULL)
...
     def package_name
-      "#{safe_base_package_name}.p5p"
+      "#{omnibus_package_name}.p5p"
+    end
+
+    #
+    # We want to have a consistent omnibus package name, but we are required
+    # to have the fmri version for the IPS repo creation.
+    #
+    def omnibus_package_name
```
schisamo: What is the purpose of introducing an extra `omnibus_package_name` method here? We should also include an iteration in the package name like we do for all other packages.

scotthain: Thanks - I missed the iteration!

The reason for having a new method is that the fmri based naming for IPS repo based packages is (for example) `chef@12.13.21,5.11-1`. It's required to name it as such for the packaging commands to work. I didn't think we wanted to name the package `chef@12.13.21,5.11-1.p5p` (tho we could). There is no 'standard' for what the p5p package looks like and usually they are shipped as like `chef.p5p`, which is obviously not going to work for us. I added the `omnibus_package_name` method to be consistent with the other omnibus package output, while still allowing for the packaging commands to function with the fmri input.

schisamo: @scotthain Shouldn't that come from the existing `fmri_package_name` method?

scotthain:
Verified locally ready for final review @chef/engineering-services

schisamo:
👍

Issue #696, opened 2016-07-28
Licensing fails when software is restored from GitCache


jkeiser:
### Description

When a software definition is restored from GitCache, and its corresponding build directory does not exist, licensing fails to find the file.

### Omnibus Version

`master`

### Platform Version

All OS's Chef builds on.

### Replication Case

This will happen anytime /var/omnibus/cache/src is empty; but one specific case where this can happen even without deleting the directory:

1. I build my project including ruby-2.1.8. /var/omnibus/cache/src/ruby/ruby-2.1.8 exists, and the licensing stuff works. ruby-2.1.8 output is in GitCache.
2. I build my project with ruby-2.2.5. GitCache doesn’t restore it, it downloads /var/omnibus/cache/src/ruby/ruby-2.2.5, and FAILS to build (for external reasons). Nothing is output to GitCache.
3. I try ruby-2.1.8 again. It restores everything from GitCache, but /var/omnibus/cache/src/ruby/ruby-2.1.8 is not there. Licensing fails.

Licensing fails because it looks for the license files in that dir, but it didn’t need to download or build the source so that dir does not exist

### Build Output

```
              [Licensing] W | 2016-07-28T21:45:05+00:00 | License file '/var/cache/omnibus/src/ruby/ruby-2.1.8/BSDL' does not exist for software 'ruby'.
              [Licensing] W | 2016-07-28T21:45:05+00:00 | License file '/var/cache/omnibus/src/ruby/ruby-2.1.8/COPYING' does not exist for software 'ruby'.
              [Licensing] W | 2016-07-28T21:45:05+00:00 | License file '/var/cache/omnibus/src/ruby/ruby-2.1.8/LEGAL' does not exist for software 'ruby'.
Encountered error(s) with project's licensing information.
Failing the build because :fatal_licensing_warnings is set in the configuration.
Error(s):

    License file '/var/cache/omnibus/src/ruby/ruby-2.1.8/BSDL' does not exist for software 'ruby'.
    License file '/var/cache/omnibus/src/ruby/ruby-2.1.8/COPYING' does not exist for software 'ruby'.
    License file '/var/cache/omnibus/src/ruby/ruby-2.1.8/LEGAL' does not exist for software 'ruby'.
```

--------------------------------------------------
/cc @chef/omnibus-maintainers

danielsdeleo:
I'm working on this as it's closely related to some work we need to do for transitive dependency license collection.

PR #700, opened 2016-08-05
Collect licenses incrementally so they cache correctly

danielsdeleo:
### Description

There are two motivations here:
1. Fix https://github.com/chef/omnibus/issues/696 by copying the license
   files from the cached source to the build directory during the
   software build (so that the license file is cached)
2. Lay the groundwork for transitive dependency license collection that
   works incrementally so that it doesn't have a similar flaw

### DIFF
--- a/lib/omnibus/software.rb
+++ b/lib/omnibus/software.rb
@@ -993,25 +993,33 @@ module Omnibus
     # be restored (if the tag does not exist), the actual build steps are
     # executed.
     #
+    # @param [Array<#execute_pre_build, #execute_post_build>] build_wrappers
+    #   Build wrappers inject behavior before or after the software is built.
+    #   They can be any object that implements `#execute_pre_build` and
+    #   `#execute_post_build`, taking this Software as an argument. Note that
+    #   these callbacks are only triggered when the software actually gets
+    #   built; if the build is skipped by the git cache, the callbacks DO NOT
+    #   run.
+    #
     # @return [true]
     #
-    def build_me
+    def build_me(build_wrappers = [])
       if Config.use_git_caching
         if project.dirty?
           log.info(log_key) do
             "Building because `#{project.culprit.name}' dirtied the cache"
           end
-          execute_build
+          execute_build(build_wrappers)
         elsif git_cache.restore
           log.info(log_key) { "Restored from cache" }
         else
           log.info(log_key) { "Could not restore from cache" }
-          execute_build
+          execute_build(build_wrappers)
           project.dirty!(self)
         end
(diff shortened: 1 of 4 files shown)


lib/omnibus/software.rb
```diff
@@ -995,23 +995,25 @@ def source_type
     #
     # @return [true]
     #
-    def build_me
+    # TODO: docco the new arg
+    # TODO: unit test that execute_pre_build and execute_post_build get called correctly
+    def build_me(build_wrappers=[])
```
danielsdeleo: I added this `build_wrapper` concept because it seemed like a good balance of being general enough to use for other things (e.g., we could make the health check incremental to take advantage of the git cache) without doing a lot of refactoring. Thoughts?

scotthain:
Looks like a great start.

sersut:
@danielsdeleo I like the `pre_build, post_build` idea. We can eventually add an api to the software dsl to expose them. That seems to be the clean way to run some logic post build, pre git_cache. 

I think this type of logic introduces a higher level question. Which option do we go with?

1. Continue to report all reports at the end.
2. Report the errors the moment we hit them (which will make the implementation a bit easier)? 

I am leaning towards (2) since it has a lower chance of ending up in whack-a-mole.

Also note that implementation of (1) might seem harder since it will require us to drop in a json file that contains the licensing information that can be git_cached to be evaluated later. But this is not actually right because we *must* drop a json file for transitive dependencies since we have the requirement to merge them and deduct the duplicate ones at the end.

Overall, I would structure the code a bit differently. I would make `Project` have an instance of `Licensing` and call `process_software(software)` on it during the post build phase of software build. This way it makes more sense to add more methods like `process_project()` `merge_licensing_data()` `clear_cache()` and still keep the control of flow in `Project` class.

danielsdeleo:
@chef/omnibus-maintainers ready for final review

@sersut - at the moment, I'd prefer to have the block-and-yield approach. It gives you one single method to call that ensures everything is in the right order. If it becomes unwieldy, I could see having methods like `#execute_pre_project` and `#execute_post_project` where you would at least mostly keep the ordering inside the `Licensing` class. But generally I think it makes sense to let the Licensing class control how it divides its work among the different methods and make sure everything gets done in the right order.

sersut:
👍

scotthain:
👍

danielsdeleo:
Running a Chef Server build as a sanity check. http://wilson.ci.chef.co/job/chef-server-12-build/375/

schisamo:
image

👍

PR #701, opened 2016-08-08
Proposal: Add an an option to include all licenses in one file

maxlazio:
### Description

We had multiple users requesting one file which contains all the generated licenses for their internal approving processes.

This PR is more of a proposal to see if this use case would be acceptable for more users, the code is written to demonstrate how this might look.

If `license_compiled_output  true`, the main `LICENSE` file would contain the actual text from each individual license, otherwise the behaviour remains the same.

My questions would be: Would something like this be useful to more omnibus users and the maintainers? If yes, how could this be improved to be accepted? If no, thanks for reading :)

### DIFF
--- a/lib/omnibus/licensing.rb
+++ b/lib/omnibus/licensing.rb
@@ -204,7 +204,8 @@ module Omnibus
     # For details, see:
     # /opt/opscode/LICENSES/python-LICENSE
     # ...
-    #
+    # If #{project.license_compiled_output} is set
+    # it will append the license contents
     # @return [String]
     #
     def components_license_summary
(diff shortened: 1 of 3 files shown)


brodock:
@chef/omnibus-maintainers is there any intention of accepting this? we are maintaining this change in our internal fork

tas50:
@maxlazio @brodock This is actually the direction I'd like to take our builds. Any chance you can rebase this so I can give it some local testing?

lamont-granquist:
obviously needs a rebase and needs fixing, but i'm cool with it.

PR #702, opened 2016-08-09
Ensure licenses dir is not removed by git cache

danielsdeleo:
### Description

Not 100% sure if this is a bug in practice or not, but any directories we create before a git cache snapshot is taken need to have a `.gitkeep` file in them or it's possible the git cache could end up removing them if a snapshot is taken when the directory is empty and later restored.


sersut:
👍

schisamo:
👍

tyler-ball:
👍

yzl:
👍

PR #704, opened 2016-08-12
Ensure compatibility of code and git cache entries

danielsdeleo:
### Description

Some code changes in omnibus can be incompatible with previously
existing git cache entries, such that builds are incorrect or fail if
the older git cache snapshot is restored in a build that is run after
upgrading omnibus. Adding a serial number to the git cache allows
omnibus to determine whether a git cache snapshot is compatible with the
current version of omnibus in use.

Note: merging this change will effectively clear the cache for every project once the project upgrades.

### DIFF
--- a/lib/omnibus/git_cache.rb
+++ b/lib/omnibus/git_cache.rb
@@ -22,6 +22,18 @@ module Omnibus
     include Util
     include Logging
 
+    # The serial number represents compatibility of a cache entry with the
+    # current version of the omnibus code base. Any time a change is made to
+    # omnibus that makes the code incompatible with any cache entries created
+    # before the code change, the serial number should be incremented.
+    #
+    # For example, if a code change generates content in the `install_dir`
+    # before cache snapshots are taken, any snapshots created before upgrade
+    # will not have the generated content, so these snapshots would be
+    # incompatible with the current omnibus codebase. Incrementing the serial
+    # number ensures these old shapshots will not be used in subsequent builds.
+    SERIAL_NUMBER = 1
+
     REQUIRED_GIT_FILES = %w{
 HEAD
 description
@@ -98,7 +110,7 @@ refs}.freeze
       # dependencies, including the on currently being acted upon.
       shasums = [dep_list.map(&:shasum), software.shasum].flatten
       suffix  = Digest::SHA256.hexdigest(shasums.join("|"))
-      @tag    = "#{software.name}-#{suffix}"
+      @tag    = "#{software.name}-#{suffix}-#{SERIAL_NUMBER}"
 
       log.internal(log_key) { "tag: #{@tag}" }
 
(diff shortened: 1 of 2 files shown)


sersut:
👍

scotthain:
A couple of questions/thoughts for discussion:
- is there an advantage to using a manually incremented serial number vs using the omnibus version or a git sha?
- Can this be set up as an opt-in/opt-out kind of feature, rather than an "always on" thing? Some teams may want to time when they do a full cache clean build?

Other than those point - looks great, thanks!

@chef/engineering-services

lib/omnibus/git_cache.rb
```diff
@@ -22,6 +22,18 @@ class GitCache
...
+    # current version of the omnibus code base. Any time a change is made to
+    # omnibus that makes the code incompatible with any cache entries created
+    # before the code change, the serial number should be incremented.
+    #
+    # For example, if a code change generates content in the `install_dir`
+    # before cache snapshots are taken, any snapshots created before upgrade
+    # will not have the generated content, so these snapshots would be
+    # incompatible with the current omnibus codebase. Incrementing the serial
+    # number ensures these old shapshots will not be used in subsequent builds.
+    SERIAL_NUMBER = 1
```
schisamo: As the git caching is a hard to understand beast, I'm worried contributors won't know the appropriate time to bump this number. Would it make more sense to base the cache busting off the `MAJOR.MINOR` of `Omnibus::VERSION `?

scotthain: I think the downside of `MAJOR.MINOR` is it does nothing for teams that ride master, unless we start releasing omnibus with every change (which is a possibility, I suppose). Would it be possible to use git sha?

danielsdeleo:
> is there an advantage to using a manually incremented serial number vs using the omnibus version or a git sha?

The idea is that you only clear the cache when you really need to. The downside is that you might forget to increment the serial number and then builds fail and you go back and do it. But many projects have really long builds and the consensus is that the tradeoffs of clearing the cache less frequently are worth it.

> Can this be set up as an opt-in/opt-out kind of feature, rather than an "always on" thing? Some teams may want to time when they do a full cache clean build?

Not really. If the code changes such that the old caches are not valid with the new code, you're just gonna have bad builds. Your options are to not upgrade omnibus, or accept that the new code is not compatible with old caches.

lib/omnibus/git_cache.rb
danielsdeleo: See my response below.

danielsdeleo:
Also, the appveyor thing is just because source forge is down. Tests pass otherwise.

yzl:
👍

lib/omnibus/git_cache.rb
danielsdeleo: I'll elaborate.

If you use `major.minor`, then:
* You can have false negatives, e.g., you can do a bug fix that looks fine but is actually incompatible with existing caches, and builds break.
* You will have false positives, e.g., a new feature (like support for a new platform) is added that doesn't affect cache validity at all, but resets your cache anyway.

If you use the SHA of the omnibus code then:
* You get a lot of false positives.
* You never get a false negative

Given the problems that clean caches can cause for us, we want to make the number of false positives as low as possible, so that makes the SHA-based option less attractive.

PR #703, opened 2016-08-10
License collection for transitive dependencies

sersut:
🚧 👷 

This branch (and PR) is being used to test Chef Client with transitive dependency licensing code. 

- [ ] Separately merge @ksubrama's changes (We need them in order to be able to test these changes with Chef Client, the first project to use this feature).
- [x] Run an ad_hoc build of Chef Client from this branch and make sure we have a successful build: http://manhattan.ci.chef.co/job/chef-trigger-ad_hoc/157/downstreambuildview/

--------------------------------------------------
/cc @chef/omnibus-maintainers

PR #705, opened 2016-08-15
Include licensing information for transitive dependencies

sersut:
This PR replaces https://github.com/chef/omnibus/pull/703 without including the changes from `ksubrama/gcc_investigate` branch.

==================

### Description

This commit adds the ability to collect licensing information from the dependencies of software definitions that are not defined in omnibus. We use license_scout gem in order to collect licensing information from the dependency managers used by the software. Since many softwares are not using dependency managers we are also including some options to enable/disable collection of licensing information:

1. New software dsl method 'skip_transitive_dependency_licensing'
When set omnibus will skip collecting licensing information for the dependencies. By default this is set to false.

2. Temporary configuration value 'fatal_transitive_dependency_licensing_warnings'
When set omnibus will fail the build when there is a warning from transitive dependency license collection. We are introducing this as a feature flag while we are implementing support for new types of dependency managers in license scout. Once we implement license detection for most commonly used set of projects, we will remove this configuration option and fall back to 'fatal_licensing_warnings'.

This feature is compatible with git_cache. We drop in a manifest and a set of license files after the software is built and before it is cached. Therefore any software restored from git_cache will always have the correct licensing information.

Just like the other licensing information, the license files of the transitive dependencies will be placed under LICENSES directory and information about them will be included in the defined license file for the project (by default LICENSE).

### Important Note

Note that we are forcing software definitions to declare `skip_transitive_dependency_licensing true` if they do not have any dependency managers. Even though this will create a lot of updates to a lot of the code, we believe this is the best approach to ensure we leak minimal amount of software without license information. See this for more information:

https://chefio.slack.com/files/dan/F1WK7KJ06/Omnibus___License_Scout_Integration_Options

### Notes

Note that this PR is not yet ready to merge. Things to do before merge: 

- [ ] Release `license_scout` gem.
- [ ] Make omnibus-software depend on the latest version of omnibus (because of the new DSL method).

/cc @chef/omnibus-maintainers

### DIFF
--- a/lib/omnibus/licensing.rb
+++ b/lib/omnibus/licensing.rb
@@ -17,6 +17,8 @@
 require "uri"
 require "fileutils"
 require "omnibus/download_helpers"
+require "license_scout/collector"
+require "license_scout/options"
 
 module Omnibus
   class Licensing
@@ -25,6 +27,7 @@ module Omnibus
     include Sugarable
 
     OUTPUT_DIRECTORY = "LICENSES".freeze
+    CACHE_DIRECTORY = "license-cache".freeze
 
     class << self
 
@@ -57,6 +60,7 @@ module Omnibus
 
           yield license_collector
 
+          license_collector.process_transitive_dependency_licensing_info
           license_collector.create_project_license_file
           license_collector.raise_if_warnings_fatal!
         end
@@ -78,12 +82,29 @@ module Omnibus
     attr_reader :licensing_warnings
 
     #
+    # The warnings encountered while preparing the licensing information for
+    # transitive dependencies.
+    #
+    # @return [Array<String>]
+    #
+    attr_reader :transitive_dependency_licensing_warnings
+
+    #
+    # Manifest data of transitive dependency licensing information
+    #
+    # @return Hash
+    #
+    attr_reader :dep_license_map
+
+    #
     # @param [Project] project
     #   the project to create licenses for.
     #
     def initialize(project)
       @project = project
       @licensing_warnings = []
+      @transitive_dependency_licensing_warnings = []
+      @dep_license_map = {}
     end
 
     #
@@ -95,6 +116,9 @@ module Omnibus
       FileUtils.rm_rf(output_dir)
       FileUtils.mkdir_p(output_dir)
       FileUtils.touch(output_dir_gitkeep_file)
+      FileUtils.rm_rf(cache_dir)
+      FileUtils.mkdir_p(cache_dir)
+      FileUtils.touch(cache_dir_gitkeep_file)
     end
 
     # Required callback to use instances of this class as a build wrapper for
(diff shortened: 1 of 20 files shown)


lib/omnibus/licensing.rb
```diff
@@ -25,6 +27,7 @@ class Licensing
     include Sugarable
 
     OUTPUT_DIRECTORY = "LICENSES".freeze
+    CACHE_DIRECTORY = "license-cache".freeze
```
danielsdeleo: Might want to put the word "license" or "licensing" in here and the related methods. I thought this was a different cache when I was reading the code.

danielsdeleo: On second thought, never mind, it should be clear from the context. I just thought I was in a different file.

lib/omnibus/licensing.rb
```diff
@@ -95,6 +116,8 @@ def prepare
       FileUtils.rm_rf(output_dir)
       FileUtils.mkdir_p(output_dir)
       FileUtils.touch(output_dir_gitkeep_file)
+      FileUtils.rm_rf(cache_dir)
+      FileUtils.mkdir_p(cache_dir)
```
danielsdeleo: Also need to touch the `cache_dir_gitkeep_file` ?

danielsdeleo: Also, I think this change will require everyone to rebuild their git caches because snapshots made before this code is merged will not have the cache dir. I'm thinking we should coordinate merging this along with https://github.com/chef/omnibus/pull/704 and do that at a time that will be least disruptive for automate.

lib/omnibus/licensing.rb
sersut: 👍

lib/omnibus/licensing.rb
sersut: 👍 Will add that. I think it dropped during the merge 😄 

Also yes. I think we should merge #704 first and then bring this in. We will need to also bump the version with this and make sure omnibus-software is pinned to the version we bump to since we will make use of `skip_transitive_dependency_licensing` there.

danielsdeleo:
👍

sersut:
The CI issues seem to be due to sourceforge connectivity.

yzl:
👍

schisamo:
Overall I find negative DSL methods like `skip_transitive_dependency_licensing` confusing. Have y'all thought about making using `transitive_dependency_licensing` which has a default value of `true`? Then you would just update software definitions where you want this disabled with `transitive_dependency_licensing false`. It might make sense to also just make the default value in this case `false` and force software definition authors to opt-in for the transitive behavior.

danielsdeleo:
We definitely want to have the behavior be to opt-out so that we minimize the chance that we accidentally miss something.

danielsdeleo:
I'm 👍  on the commits since my last 👍 Are the test failures all issues with that source forge URL? Might want to fix that soon....

PR #709, opened 2016-08-25
Update chefstyle and fix issues

scotthain:
### Description

Briefly describe the new feature or fix here

chefstyle 0.4.0 came out, with updated rules.

--------------------------------------------------
/cc @chef/omnibus-maintainers <- This ensures the Omnibus Maintainers team are notified to review this PR.

#### Maintainers

Please ensure that you check for:

- [] If this change impacts git cache validity, it bumps the git cache
  serial number
- [] If this change impacts compatibility with omnibus-software, the
  corresponding change is reviewed and there is a release plan
- [] If this change impacts compatibility with the omnibus cookbook, the
  corresponding change is reviewed and there is a release plan

@mwrock resolves the style violations in https://github.com/chef/omnibus/pull/708

please take an extra look at the 3 manual changes I made that didn't autocorrect:

```
lib/omnibus/core_extensions/open_uri.rb:69:5: C: Style/ConstantName: Use SCREAMING_SNAKE_CASE for constants.
    StringMax = 0
    ^^^^^^^^^
lib/omnibus/fetchers/git_fetcher.rb:217:5: W: Lint/UselessAccessModifier: Useless public access modifier.
    public
    ^^^^^^
lib/omnibus/manifest.rb:131:5: W: Lint/UselessAccessModifier: Useless private access modifier.
    private
    ^^^^^^^
```

@chef/engineering-services

PR #710, opened 2016-08-25
Transitive dependency licensing fixes

sersut:
### Description

This PR includes multiple fixes around transitive dependency licensing. Please see the individual commit messages for details. 

- [x] Make sure the private branch in the Gemfile gets removed before merge.

/cc: @danielsdeleo @ryancragun 

/cc @chef/omnibus-maintainers

### DIFF
--- a/lib/omnibus/git_cache.rb
+++ b/lib/omnibus/git_cache.rb
@@ -32,7 +32,7 @@ module Omnibus
     # will not have the generated content, so these snapshots would be
     # incompatible with the current omnibus codebase. Incrementing the serial
     # number ensures these old shapshots will not be used in subsequent builds.
-    SERIAL_NUMBER = 1
+    SERIAL_NUMBER = 2
 
     REQUIRED_GIT_FILES = %w{
 HEAD
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,7 @@ gemspec
 gem "pedump", git: "https://github.com/ksubrama/pedump", branch: "patch-1"
 
 # Always use license_scout from master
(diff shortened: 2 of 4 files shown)


schisamo:
@sersut I think you'll need fixes in https://github.com/chef/omnibus/pull/709, we should get that merged and you can just rebase on master.

scotthain:
@sersut go ahead and rebase and run through travis again

lib/omnibus/git_cache.rb
```diff
@@ -32,7 +32,7 @@ class GitCache
     # will not have the generated content, so these snapshots would be
     # incompatible with the current omnibus codebase. Incrementing the serial
     # number ensures these old shapshots will not be used in subsequent builds.
-    SERIAL_NUMBER = 1
+    SERIAL_NUMBER = 2
```
danielsdeleo: Is this because there could be cache snapshots with license collection errors in them?

danielsdeleo:
👍 changes look good. Just remember to do the cache rebuild dance with automate, etc. if we really have to bump the cache serial number.

lib/omnibus/git_cache.rb
sersut: There are a few reasons:

1. Cache might not contain the license info collected by `chef_berkshelf` plugin. Technically anytime we add a new dep mgr, we need to bump this. 
2. The name of the project we're passing has changed which requires us to invalidate the old caches.

sersut:
Hrmm looks like rebar license collection has an issue in it. I will fix that before merging this since that will also require a `SERIAL_NUMBER` bump.

PR #726, opened 2016-10-22
Add packager for FreeBSD (pkgng)

amdprophet:
### Description

Adds a new packager for FreeBSD (pkgng).

---

#### Maintainers

  serial number

### DIFF
--- a/lib/omnibus/packager.rb
+++ b/lib/omnibus/packager.rb
@@ -30,6 +30,7 @@ module Omnibus
     autoload :Solaris,  "omnibus/packagers/solaris"
     autoload :IPS,      "omnibus/packagers/ips"
     autoload :RPM,      "omnibus/packagers/rpm"
+    autoload :PKGNG,    "omnibus/packagers/pkgng"
 
     #
     # The list of Ohai platform families mapped to the respective packager
@@ -48,6 +49,7 @@ module Omnibus
       "ips"      => IPS,
       "windows"  => [MSI, APPX],
       "mac_os_x" => PKG,
+      "freebsd"  => [Makeself, PKGNG],
       "smartos"  => PKGSRC,
     }.freeze
 
--- a/lib/omnibus/packagers/pkgng.rb
+++ b/lib/omnibus/packagers/pkgng.rb
@@ -1,0 +1,571 @@
+require "ffi_yajl"
+
+module Omnibus
+  class Packager::PKGNG < Packager::Base
+    # @return [Hash]
+    SCRIPT_MAP = {
+      # Default Omnibus naming
+      preinst:  "pre-install",
+      postinst: "post-install",
+      inst:     "install",
+      prerm:    "pre-deinstall",
+      postrm:   "post-deinstall",
+      rm:       "deinstall",
+      preup:    "pre-upgrade",
+      postup:   "post-upgrade",
+      up:       "upgrade",
+    }.freeze
+
+    id :pkgng
+
+    setup do
+      # Copy the full-stack installer into staging_dir, accounting for
+      # any excluded files.
+      #
+      # /opt/halmet => /tmp/daj29013/opt/hamlet
+      destination = File.join(staging_dir, project.install_dir)
+      FileSyncer.sync(project.install_dir, destination, exclude: exclusions)
+
+      # Copy over any user-specified extra package files.
+      #
+      # Files retain their relative paths inside the scratch directory, so
+      # we need to grab the dirname of the file, create that directory, and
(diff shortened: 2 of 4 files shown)


lib/omnibus/packagers/pkgng.rb
```diff
@@ -0,0 +1,574 @@
...
+    #     }
+    #   }
+    #
+    # @param [Hash] val
+    #   the dependencies for this package
+    #
+    # @return [Hash]
+    #   the dependencies for this package
+    #
+    def dependencies(val = NULL)
```
yzl: If this is for runtime dependencies, I'd rather call it `runtime_dependencies`; the `omnibus` DSL uses `dependency` and `dependencies` for buildtime dependencies.

amdprophet: I've renamed this to `runtime_dependencies` in the latest commit.

schisamo:
@amdprophet Please squash everything down into a single commit and ensure it is signed off per the [Chef DCO](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

amdprophet:
@schisamo squashed and signed off, thanks!

schisamo:
@chef/client-freebsd I would like y'all to give this a review also.

lib/omnibus/packager.rb
```diff
@@ -48,6 +49,7 @@ module Packager
       "ips"      => IPS,
       "windows"  => [MSI, APPX],
       "mac_os_x" => PKG,
+      "freebsd"  => PKGNG,
```
schisamo: To ensure some backward compat let's produce both a `Makeself` and `PKGNG` artifact on FreeBSD. The following code should work (be sure to update the specs also):

```ruby
"freebsd"  => [Makeself, PKGNG],
```

lib/omnibus/packagers/pkgng.rb
```diff
@@ -0,0 +1,574 @@
...
+      File.open(destination, "w+") do |f|
+        FFI_Yajl::Encoder.encode(compact_manifest.merge(manifest), pretty: true).each_line do |line|
+          f.write(line)
+        end
+      end
+    end
+
+    #
+    # Create the +.txz+ file, compressing at xz level 9. The use of the
+    # +fakeroot+ command is required so that the package is owned by
```
srenatus: Is this comment still accurate? I don't see any `fakeroot` invocation there... But I could be missing something.

lib/omnibus/packagers/pkgng.rb
```diff
@@ -0,0 +1,574 @@
...
+      destination = File.join(staging_dir, "+MANIFEST")
+      File.open(destination, "w+") do |f|
+        FFI_Yajl::Encoder.encode(compact_manifest.merge(manifest), pretty: true).each_line do |line|
+          f.write(line)
+        end
+      end
+    end
+
+    #
+    # Create the +.txz+ file, compressing at xz level 9. The use of the
```
srenatus: Would we want to make this configurable? If FreeBSD is commonly using xz level 9, I suppose we could, too. However, the xz command line utilities warn about level 7-9:

```
  -0 ... -9           compression preset; default is 6; take compressor *and*
                      decompressor memory usage into account before using 7-9!
```

I couldn't find quickly what `pkg create` would end up using...

lib/omnibus/packagers/pkgng.rb
scotthain: Guessing it's a copypasta from https://github.com/chef/omnibus/blob/master/lib/omnibus/packagers/deb.rb#L363-L365 ?

amdprophet: It was indeed copypasta. I'm not sure on a default for `pkg create` off the top of my head. I'll look into it.

lib/omnibus/packagers/pkgng.rb
amdprophet: I mistakingly commented that `pkg create` was using a default xz compression level of 9. The man pages for `xz` shows that the default is 6 which should be what's being used here.

scotthain:
Looks like we need to add a `pkg_safe` sort of method (similar to the `rpm_safe` method here: https://github.com/chef/omnibus/blob/master/lib/omnibus/packagers/rpm.rb#L526-L534) to ensure that if any files have restricted characters, they are substituted. I'm also running into a strange issue with symlinks not picking up correctly, but I think that may be my test setup.

amdprophet:
I'm going to close this as we aren't using this project anymore.

PR #728, opened 2016-10-24
[BEN-123] Fix IPS package must create symlinks to package commands

jaymalasinha:
### Description

IPS package creation script looks for symlinks.erb in local project resources and adds them to
manifest file so the IPS package will install symlinks from the template. These are the links added by post install scripts for other package types. But since IPS does not use post install scripts they need to be added here.

Fixes issue: https://github.com/chef/chef/issues/5458
---

@chef/engineering-services 
#### Maintainers

  serial number
  - Looks for symlinks.erb in project directories and adds them to
    manifest file
  - Add unit tests for symlinks file creation
  - Update unit tests for manifest file creation

Signed-off-by: Jaymala Sinha jsinha@chef.io

### DIFF
--- a/lib/omnibus/packagers/ips.rb
+++ b/lib/omnibus/packagers/ips.rb
@@ -193,6 +193,24 @@ module Omnibus
     end
 
     #
+    # A set of symbolic links to installed commands that
+    #`pkgmogrify' will apply to the package manifest. Is called only when
+    # symlinks.erb template exists
+    # The resource exists locally. For example for project omnibus-toolchain
+    # resource_path("symlinks.erb") #=>
+    # {"/path/to/omnibus-toolchain/resources/omnibus-toolchain/ips/symlinks.erb"}
+    #
+    # @return [String]
+    #
+    def render_symlinks
+      render_template_content(resource_path("symlinks.erb"),
+        {
+          projectdir: project.install_dir,
+        }
+      )
+    end
+
+    #
     # Generate package metadata
     #
     # Create the gen template for `pkgmogrify`
@@ -210,6 +228,13 @@ module Omnibus
           arch:              safe_architecture,
         }
       )
+
+      # Append the contents of symlinks.erb if it exists
+      if File.exists?(resource_path("symlinks.erb"))
+        File.open(pkg_metadata_file, "a") do |symlink|
+          symlink.write(render_symlinks)
+        end
+      end
 
       # Print the full contents of the rendered template file to generate package contents
       log.debug(log_key) { "Rendered Template:\n" + File.read(pkg_metadata_file) }
--- a/lib/omnibus/templating.rb
+++ b/lib/omnibus/templating.rb
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-
 require "erb"
 
 module Omnibus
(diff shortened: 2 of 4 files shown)


lib/omnibus/packagers/ips.rb
```diff
@@ -211,6 +239,15 @@ def write_pkg_metadata
         }
       )
 
+      # Append the contents of symlinks.erb if it exists
+      if File.exists?(resource_path("symlinks.erb"))
+        write_symlinks_file
+        add_symlinks = File.read(symlinks_file)
```
schisamo: It would be more efficient to not write to a file and turn around and read from it. Perhaps we can add a new method to the [`Omnibus::Templating` mixin](https://github.com/chef/omnibus/blob/master/lib/omnibus/templating.rb) that can just render some ERB to a string? You could pull the non-file writing functionality out of `Omnibus::Templating::render_template` into a separate method (in the same module) that just returns the rendered content as a string and update `render_template` to use the new method.

jaymalasinha: @schisamo updated templating

lib/omnibus/templating.rb
```diff
@@ -39,14 +38,9 @@ def self.included(base)
     # @option options [Hash] :variables (default: +{}+)
     #   the list of variables to pass to the template
     #
-    def render_template(source, options = {})
-      destination = options.delete(:destination) || source.chomp(".erb")
-
-      mode      = options.delete(:mode) || 0644
+    def render_template_var(source, options)
```
schisamo: Let's call this `render_template_content`.

lib/omnibus/templating.rb
```diff
@@ -61,10 +55,22 @@ def render_template(source, options = {})
           Struct.new(*variables.keys).new(*variables.values)
         end
 
-      result = template.result(struct.instance_eval { binding })
+      @result = template.result(struct.instance_eval { binding })
+      @result
```
schisamo: This method should just return the result of `template.result` directly instead of setting that value to an instance variable (`@result`) and turning around and returning the instance variable.

lib/omnibus/templating.rb
```diff
@@ -61,10 +55,22 @@ def render_template(source, options = {})
...
+
+    def render_template(source, options = {})
+      @source = source
+      @options = options
+
+      destination = options.delete(:destination) || source.chomp(".erb")
+      mode = options.delete(:mode) || 0644
+      log.info(log_key) { "Rendering `#{source}' to `#{destination}'" }
+
+      render_template_var(source, options)
```
schisamo: This should be:

```
result = render_template_var(source, options)
```

lib/omnibus/templating.rb
```diff
@@ -61,10 +55,22 @@ def render_template(source, options = {})
...
 
       File.open(destination, "w", mode) do |file|
-        file.write(result)
+        file.write(@result)
```
schisamo: Assuming you return the result above then this should be:

```
file.write(result)
```

jaymalasinha:
@schisamo Updated as per comments.

spec/functional/templating_spec.rb
```diff
@@ -59,7 +59,9 @@ class RandomClass
           expect(destination).to be_an_executable
         end
       end
+    end
 
+    describe "#render_template_content" do
```
schisamo: It looks like this test is calling `render_template` (on line 80) instead of `render_template_content`.

jaymalasinha:
@schisamo fixed the tests and verified the build working.

PR #732, opened 2016-11-22
Add some more file endings and path patterns to exclude

srenatus:
### Description

This gets the health check time of chef-server from 44s to 27s. (Haven't had a chance to test it with Automate yet, but this is where I draw inspiration from for further stuff to exclude...)

For future reference: changing the `find` call to 
```
read_shared_libs("find #{project.install_dir}/ -type f -regextype posix-extended ! -regex '#{regexp}' -exec sh -c \"readelf -d {} >/dev/null 2>&1\" \\; -print0 | xargs -0 ldd") do |line|
```

to exclude further files -- _by paying with a shellout of `sh` and `readelf` to figure out if that's save_ --saves another 4s only. Since this also requires readelf (package binutils) to be present, I refrain from proposing that change (for now...)

#### Maintainers

- [X] If this change impacts git cache validity, it bumps the git cache
  serial number
- [X] If this change impacts compatibility with omnibus-software, the
- [X] If this change impacts compatibility with the omnibus cookbook, the

### DIFF
--- a/lib/omnibus/health_check.rb
+++ b/lib/omnibus/health_check.rb
@@ -159,29 +159,38 @@ module Omnibus
     ].freeze
 
     IGNORED_ENDINGS = %w{
+      .TXT
       .[ch]
+      .bat
+      .beam
+      .cmake
       .e*rb
       .gemspec
       .gitignore
       .h*h
       .java
       .js
+      .jsm
       .json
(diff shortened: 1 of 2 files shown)


srenatus:
Nobody move, appveyor's green. 😉

PR #733, opened 2016-11-25
Build pkgsrc packages

### LINKED ISSUE #321: SmartOS can't install SVR4-style Solaris packages
Let me try this again in the right repo...

SmartOS does not have the old Solaris `pkgadd` commands as it uses BSD pkgsrc under the covers. Thus, our attempt to coerce it into installing a SVR4-type .pkg file is actually incorrect.

If we actually do intend to support SmartOS in the context of the platform support RFC, then we would need to build native pkgsrc packages for it. (You could `pkgtrans` the Solaris ones to it as part of CI or something, as a cheap trick.)

thommay:
For now, only tested on SmartOS. Fixes #321 

/cc @chef/omnibus-maintainers @juliandunn

### DIFF
--- a/lib/omnibus/health_check.rb
+++ b/lib/omnibus/health_check.rb
@@ -124,6 +124,12 @@ module Omnibus
       /libgcc_s.so/,
       /libstdc\+\+\.so/,
       /libcrypt.so/,
+      /libcrypto.so/,
+      /libssl.so/,
+      /libssp\.so/,
+      /libumem\.so/,
+      /libffi\.so/,
+      /libz\.so/, # while we package our own libz, this get dragged along from Solaris 11's libelf library for some reason...
     ].freeze
 
     MAC_WHITELIST_LIBS = [
--- a/lib/omnibus/packager.rb
+++ b/lib/omnibus/packager.rb
@@ -26,6 +26,7 @@ module Omnibus
     autoload :MSI,      "omnibus/packagers/msi"
     autoload :APPX,     "omnibus/packagers/appx"
     autoload :PKG,      "omnibus/packagers/pkg"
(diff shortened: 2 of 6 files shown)


thommay:
(I can't restart appveyor but it's SF timeout related)

lib/omnibus/health_check.rb
```diff
@@ -124,6 +124,12 @@ class HealthCheck
       /libgcc_s.so/,
       /libstdc\+\+\.so/,
       /libcrypt.so/,
+      /libcrypto.so/,
+      /libssl.so/,
```
stevendanna: These are a bit worrying.  Shouldn't these always be safely provided by OpenSSL?

scotthain: These are an unfortunate side-effect of Solaris 11's Cryptographic Framework, which kind of includes ssl in the kernel. If you're morbidly curious: https://docs.oracle.com/cd/E23824_01/html/821-1456/scf-10.html#scrolltoc

thommay:
Actually for reasons I don't understand, there seems to be something wonky with SSL:
```
[root@motet ~]# knife ssl check -c /etc/chef/client.rb
Connecting to host api.chef.io:443
ERROR: The SSL certificate of api.chef.io could not be verified
Certificate issuer data: /C=US/O=DigiCert Inc/CN=DigiCert SHA2 Secure Server CA
```

like it's not using the omnibus trust store?

scotthain:
@thommay I've never seen that one before... I might suggest commenting out all the chef defs in your project file and add a curl dep, so you can debug the SSL through curl. I've had some success with that in the past.

thommay:
The ssl thing appears transient, so i'm gonna merge

PR #735, opened 2016-11-28
Remove version constrain on chefstyle

tas50:
We should let chefstyle float


lamont-granquist:
as long as you understand that i'm not signing up to submit PRs against all the downstream repos every time i do a release of chefstyle...   so i'll do my roughly quarterly chefstyle release and then travis will go red here...

tas50:
If we think it's our style guide we should be prepared to fix it as we do releases of the various projects. Otherwise we have multiple different versions of the style guide implemented across projects as we slowly upgrade everything to the latest release.

lamont-granquist:
as long as the pain of upgrades are decentralized...

PR #736, opened 2016-12-06
Add support for S3 transfer acceleration for S3 Cache

brodock:
### Description

Omnibus can use S3 to cache software assets used during the build to speedup the process.

If we have a slow connection to S3, it will slow down the whole build process. To fight that we can either enable S3 Transfer Acceleration, or host something like "MinIO" closer to where the build runs.

Unfortunately, there is no way to either change the endpoint or enable the S3 transfer acceleration flag to the `aws-sdk` in omnibus.

From the `NetFetcher`, the endpoint is hardcoded (you can only customize the bucket name):
> https://github.com/chef/omnibus/blob/3eefb1cd8de69b1d97de2962f779512892e9296d/lib/omnibus/fetchers/net_fetcher.rb#L145-L151

And From the `S3Cache` you can't set the `:endpoint` [configuration param](https://github.com/aws/aws-sdk-ruby#configuration) nor can you define the `:use_accelerate_endpoint` [flag](https://github.com/aws/aws-sdk-ruby/blob/9e05bc17ad7a20379820301c5d21692fff9f99dd/aws-sdk-core/lib/aws-sdk-core/plugins/s3_accelerate.rb#L10):
> https://github.com/chef/omnibus/blob/3eefb1cd8de69b1d97de2962f779512892e9296d/lib/omnibus/s3_cache.rb#L137-L144

## Proposal

Make `:endpoint` configurable in S3Cache, and use it to generate the URL in `NetFetcher`.
the new configuration in `omnibus.rb` configuration file will be called: `s3_endpoint`

Add a configuration flag to optionally use the S3 Transfer Acceleration within the AWS-SDK bundled support.

Both solutions are required to enable either custom S3 endpoints, or to use the AWS officially supported S3 Transfer Acceleration.

#### Maintainers

  serial number

### DIFF
--- a/lib/omnibus/config.rb
+++ b/lib/omnibus/config.rb
@@ -297,6 +297,16 @@ module Omnibus
       "us-east-1"
     end
 
+    # The HTTP or HTTPS endpoint to send requests to, when using non-standard endpoint
+    #
+    # @return [String, nil]
+    default(:s3_endpoint, nil)
+
+    # Enable or disable S3 Accelerate support
+    #
+    # @return [true, false]
+    default(:s3_accelerate, false)
+
     # --------------------------------------------------
     # @!endgroup
     #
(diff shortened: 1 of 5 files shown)


brodock:
Rebased :)

stevendanna:
gif-keyboard-13311758052196882916

PR #737, opened 2016-12-08
Add msys2 support behind feature flag

tduffield:
### Description

This takes the important bits from the existing msys2 branch and hides certain portion of the code behind a feature flag controlled by the `MSYS_TOOLCHAIN` environment flag. When the omnibus-toolchain modifications are complete, we should be able to come back through and remove these components. 

#### Maintainers

  serial number

### DIFF
--- a/lib/omnibus/util.rb
+++ b/lib/omnibus/util.rb
@@ -111,8 +111,10 @@ module Omnibus
       # chef-client/chef-dk installations from the path.
       if options.delete(:clean_ruby_path)
         path_dirs = options[:environment].fetch(path_key, "").split(File::PATH_SEPARATOR || ":")
+        path_dirs = path_dirs.map { |p| windows_safe_path(p) }
+        safe_install_dir = windows_safe_path(install_dir)
         path_dirs = path_dirs.reject do |p|
-          filter_paths.any? { |f| windows_safe_path(p).start_with?(f) }
+          !p.start_with?(safe_install_dir) && filter_paths.any? { |f| safe_p.start_with?(f) }
         end
         options[:environment][path_key] = path_dirs.join(File::PATH_SEPARATOR || ":")
       end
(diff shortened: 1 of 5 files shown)


stevendanna:
Even though it passes the bot, we should probably add sign-offs to all the commits.

tduffield:
@stevendanna done!

tduffield:
@schisamo I'm happy to make it an Omnibus config if that's what you prefer. The goal here is to separate the work to get omnibus-toolchain working with "de-risking" the fact that chef and chef-dk are being built using long-running branches. Getting these PRs merged into master frees us up to prioritize the omnibus-toolchain work independently of the risk associated with those long-running branches.

lib/omnibus/util.rb
```diff
@@ -65,18 +99,56 @@ def shellout(*args)
...
+      path_keys = options[:environment].keys.grep(/\Apath\Z/i)
+      if path_keys.length > 1
+        raise InvalidValue.new("shellout", "receive environment without duplicate PATH values")
+      elsif path_keys.length == 1 && path_keys.first != path_key
+        options[:environment][path_key] = options[:environment].delete(path_keys.first)
+      end
+
+      # Try our best to get rid of the "outer" omnibus ruby and any current
+      # chef-client/chef-dk installations from the path.
+      if ENV["MSYS_TOOLCHAIN"]
```
stevendanna: In other places we do `windows? && ENV["MSYS_TOOLCHAIN"]`.  Any reason to exclude the windows? check here?  Perhaps we could have a use_msys_toolchain? function that does both of those checks always.

lib/omnibus/util.rb
```diff
@@ -65,18 +99,56 @@ def shellout(*args)
         options[:environment] = options.fetch(:environment, {}).merge(options[:env])
       end
 
+      # Double check that we don't have conflicting PATH definitions.
+      path_keys = options[:environment].keys.grep(/\Apath\Z/i)
+      if path_keys.length > 1
+        raise InvalidValue.new("shellout", "receive environment without duplicate PATH values")
```
stevendanna: Should this be "with" rather than "without"?

lib/omnibus/util.rb
```diff
@@ -65,18 +99,56 @@ def shellout(*args)
...
+      # chef-client/chef-dk installations from the path.
+      if ENV["MSYS_TOOLCHAIN"]
+        path_dirs = options[:environment].fetch(path_key, "").split(File::PATH_SEPARATOR || ":")
+        path_dirs = path_dirs.map { |p| windows_safe_path(p) }
+        safe_install_dir = windows_safe_path(install_dir)
+        path_dirs = path_dirs.reject do |p|
+          !p.start_with?(safe_install_dir) && filter_paths.any? { |f| p.start_with?(f) }
+        end
+        options[:environment][path_key] = path_dirs.join(File::PATH_SEPARATOR || ":")
+      end
```
stevendanna: It might be nice to move the contents of this `if` block into a function that takes an environment and returns a new environment.

PR #738, opened 2016-12-08
Make RPM and DEB package compression configurable

### LINKED ISSUE #422: Request: Configurable dpkg-deb compression args
I noticed that the deb packager (lib/omnibus/packagers/deb.rb) forces a particular set of dpkg-deb compression args, specifically '-z9 -Zgzip'.

This is nice in one sense, because it gives a shorter run time for dpkg-deb. Unfortunately, it leaves me with rather large .deb files.

In contrast, dpkg-deb's default compression values are '-z6 -Zxz', which leads to ~2x run times but ~0.5x deb size (in my case).

Falling back to dpkg-deb's default values seems unnecessary as it would lead to annoying build times, so keeping a faster compression setting as the Omnibus default seems like a good idea. But once I'm done tinkering with the project, being able to specify a slower but stronger compression mechanism would be great for building a smaller and moar-faster-at-deploy-time package.

srenatus:
### Description

This allows for configuring the type and level of compression to use for DEB and RPM packages.

It adds new configurables for `config/projects/xyz.rb`, defaulting to what was hardcoded before. For example, choosing xz:

```ruby
package :deb do
  compression_level 6
  compression_type :xz
  compression_strategy :extreme # /!\ dpkg-deb 1.16.1.2 doesn't seem to know this
end
package :rpm do
  compression_level 6
  compression_type :xz
end
```

Fixes #422.

#### Maintainers

  serial number

### DIFF
--- a/lib/omnibus/packagers/deb.rb
+++ b/lib/omnibus/packagers/deb.rb
@@ -167,6 +167,86 @@ module Omnibus
     expose :section
 
     #
+    # Compression algorithm (gzip, xz, none) to use (-Z).
+    #
+    # @example
+    #   compression_type :xz
+    #
+    # @param [Symbol] val
+    #   type of compression (:gzip, :xz, :none)
+    #
+    # @return [Symbol]
+    #   type of compression for this package
+    #
+    def compression_type(val = NULL)
+      if null?(val)
+        @compression_type || :gzip
+      else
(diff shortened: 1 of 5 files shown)


srenatus:
@schisamo thanks for the pointers, I wasn't aware of this at all :)

srenatus:
@schisamo I've changed it 🎉 

Appveyor isn't well-connected, though. I'll go on re-running this.

PR #745, opened 2017-01-18
[BEN-127] Add version-lock facet at end of manifest

jaymalasinha:
When version-lock facet to remove package version
dependencies is added to the doc-transform file it gets overwritten
while resolving dependencies.  So adding it right before publishing
the manifest

### Description

The package created from this fix was tested on Solaris11.2.  I was able to install the generated package after setting the version-lock to false at the command line with
# pkg change-facet facet.version-lock.*=false

#### Maintainers

  serial number


jaymalasinha:
@chef/engineering-services

PR #746, opened 2017-01-24
[BEN-147] Add omnibus-gcc libraries to LD_RUN_PATH for s390x and RHEL6

jaymalasinha:

### Description

@chef/engineering-services 
#### Maintainers

  serial number

### DIFF
--- a/lib/omnibus/software.rb
+++ b/lib/omnibus/software.rb
@@ -754,6 +754,13 @@ module Omnibus
         "LD_RUN_PATH" => "#{install_dir}/embedded/lib",
       }
 
+      if s390x? && platform_version.satisfies?("< 7.0")
+        extra_linker_flags = {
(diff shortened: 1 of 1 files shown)


lib/omnibus/software.rb
```diff
@@ -754,6 +754,13 @@ def with_standard_compiler_flags(env = {}, opts = {})
         "LD_RUN_PATH" => "#{install_dir}/embedded/lib",
       }
 
+      if s390x? && platform_version.satisfies?("<= 6.8")
```
scotthain: We should make this match any 6.x versions, in case there is a 6.9+

jaymalasinha:
@scotthain good point. updated as per comments

jaymalasinha:
Was updated through opscode-ci since the oc_id uses system libraries, this change will not take effect.

PR #752, opened 2017-02-06
Fix appx packager

rhass:
### Description

This fixes two issues with building appx packages.

- Translate string quotes `"` from return value of `certificate_subject` method to XML `&quot;` for insertion in ERB XML template.
- Upgrades default signing algorithm from SHA1 to SHA256 for both MSI and APPX. SHA256 is the minimum allowed signing algorithm for appx packages, and SHA1 is no longer a recommended algorithm.

#### Maintainers

- [ ] If this change impacts git cache validity, it bumps the git cache serial number
- [ ] If this change impacts compatibility with omnibus-software, the corresponding change is reviewed and there is a release plan.
- [ ] If this change impacts compatibility with the omnibus cookbook, the corresponding change is reviewed and there is a release plan.

### DIFF
--- a/lib/omnibus/packagers/appx.rb
+++ b/lib/omnibus/packagers/appx.rb
@@ -68,7 +68,7 @@ module Omnibus
           friendly_name:   project.friendly_name,
           version:         windows_package_version,
           maintainer:      project.maintainer,
-          certificate_subject: certificate_subject,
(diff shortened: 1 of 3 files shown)


rhass:
Updated PR to fix chefstyle issue.

PR #753, opened 2017-02-10
Retry git network operations by default

stevendanna:
We see occasional build failures because of transient network issues
between the build environment and upstream git providers. To reduce
the number of failures caused by transient outages, this provides a
simple retry helper that we then use to wrap the network operations in
the git fetcher.

#### Maintainers

  serial number

### DIFF
--- a/lib/omnibus/fetchers/git_fetcher.rb
+++ b/lib/omnibus/fetchers/git_fetcher.rb
@@ -141,7 +141,9 @@ module Omnibus
     # @return [void]
     #
     def git_clone
-      git("clone#{" --recursive" if clone_submodules?} #{source_url} .")
+      retry_block("git clone", [CommandTimeout, CommandFailed]) do
+        git("clone#{" --recursive" if clone_submodules?} #{source_url} .")
+      end
     end
 
     #
@@ -154,7 +156,11 @@ module Omnibus
       # default when a sha1 is provided).  git older than 1.7.5 doesn't
       # support the --detach flag.
       git("checkout #{resolved_version} -f -q")
-      git("submodule update --recursive") if clone_submodules?
+      if clone_submodules?
+        retry_block("git submodule update", [CommandTimeout, CommandFailed]) do
+          git("submodule update --recursive")
+        end
+      end
     end
 
     #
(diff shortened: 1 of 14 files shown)


lib/omnibus/fetchers/git_fetcher.rb
```diff
@@ -165,7 +171,9 @@ def git_checkout
     def git_fetch
       fetch_cmd = "fetch #{source_url} #{described_version}"
       fetch_cmd << " --recurse-submodules=on-demand" if clone_submodules?
-      git(fetch_cmd)
+      retry_block("git fetch", [CommandTimeout, CommandFailed]) do
```
schisamo: It might be nice to know the full fetch command that times out...maybe we can do this:

```ruby
retry_block(fetch_cmd, [CommandTimeout, CommandFailed]) do
```

lib/omnibus/fetchers/git_fetcher.rb
```diff
@@ -259,7 +267,10 @@ def self.revision_from_remote_reference(ref, source)
       # allows us to return the SHA of the tagged commit for annotated
       # tags. We take care to only return exact matches in
       # process_remote_list.
-      remote_list = shellout!("git ls-remote \"#{source[:git]}\" \"#{ref}*\"").stdout
+      remote_list = retry_block("git ls-remote", [CommandTimeout, CommandFailed]) do
```
schisamo: Similar to my comment above...knowing the source would that `git ls-remote` is calling would be helpful for debug situations.

lib/omnibus/fetchers/git_fetcher.rb
stevendanna: @schisamo I can certainly add more info, the only reason I didn't is because if it fails the retry, then you get the full pretty-printing of the exception which has a lot of info.  This is just for the log message telling you it was retried.  I am happy with adding more info though if you think it'll be helpful.

schisamo: Nope sounds like the data we need would be in that pretty output. :shipit:!

PR #759, opened 2017-03-02
Add support for building against omnibus-toolchain on Windows

rhass:
### Description

This cleans up the build process for Windows by adding supporting code to use `omnibus-toolchain` on Windows and adds compiler optimizations. Moreover, this removes a large chunk of code which is no longer needed to build Windows projects.

*Important Note:* _Chef, ChefDK, Inspec, and omnibus-toolchain are still pinned to the existing crazy long-standing branches, and will need to remain pinned until we have replaced the internal CI build and test systems for each corresponding project._

/cc @chef/omnibus-maintainers 
/cc @chef/engineering-services 
/cc @tduffield @ksubrama

#### Maintainers

- [x] This change impacts git cache validity
  - [x]  The git cache serial number has been correctly bumped
- [x] This change impacts compatibility with omnibus-software
  - [x] The corresponding change is reviewed
  - [x] There is a release plan
- [x] If this change impacts compatibility with the omnibus cookbook
  - [x] The corresponding change is reviewed 
  - [x] There is a release plan

### DIFF
--- a/lib/omnibus/git_cache.rb
+++ b/lib/omnibus/git_cache.rb
@@ -32,7 +32,7 @@ module Omnibus
     # will not have the generated content, so these snapshots would be
     # incompatible with the current omnibus codebase. Incrementing the serial
     # number ensures these old shapshots will not be used in subsequent builds.
-    SERIAL_NUMBER = 2
+    SERIAL_NUMBER = 3
 
     REQUIRED_GIT_FILES = %w{
 HEAD
--- a/lib/omnibus/builder.rb
+++ b/lib/omnibus/builder.rb
@@ -276,6 +276,21 @@ module Omnibus
     expose :windows_safe_path
 
     #
+    # (see Util#compiler_safe_path)
+    #
+    # Some compilers require paths to be formatted in certain ways. This helper
+    # takes in the standard Omnibus-style path and ensures that it is passed
+    # correctly.
+    #
+    # @example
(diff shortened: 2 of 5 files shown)


rhass:
It looks like the duplicate path test failure is legit. I will need to look at this in the AM, but if anyone has some insight on this issue it would be appreciated. Thanks!

rhass:
I am guessing we should probably collapse this all too before we merge.

rhass:
The integration tests are failing due toSourceForge issues.

lib/omnibus/git_cache.rb
```diff
@@ -32,7 +32,7 @@ class GitCache
     # will not have the generated content, so these snapshots would be
     # incompatible with the current omnibus codebase. Incrementing the serial
     # number ensures these old shapshots will not be used in subsequent builds.
-    SERIAL_NUMBER = 2
+    SERIAL_NUMBER = 3
```
schisamo: Does this need to be bumped?

scotthain: This does need to be bumped because the an existing cached build will have unpredictable results since the flags/compilation steps have changed. It's safer to force a clear.

rhass:
@schisamo I collapsed the changes, we should be good to go.

lib/omnibus/git_cache.rb
schisamo: Makes sense.

PR #766, opened 2017-03-20
Ensure that Windows Paths are correctly frontwhacked

thommay:
Dir.glob only handles forward slashes in paths, so mixed or backwhacked
paths return nothing.

cc @chef/omnibus-maintainers

### DIFF
--- a/lib/omnibus/file_syncer.rb
+++ b/lib/omnibus/file_syncer.rb
@@ -34,7 +34,7 @@ module Omnibus
     #   the list of all files
     #
     def glob(pattern)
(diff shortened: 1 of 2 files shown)


thommay:
by way of an example, `glob('C:\opscode\angry-omnibus-toolchain/embedded/bin/mingw64/bin/libwinpthread-1.dll')` won't match `C:/opscode/angry-omnibus-toolchain/embedded/bin/mingw64/bin/libwinpthread-1.dll` with the current glob implementation

schisamo:
@thommay Can we add some test coverage for this change? We'll also need a rebase onto master.

PR #772, opened 2017-04-11
Add Chef::Sugar::Init to Sugarable DSL

legal90:
### Description

This PR adds [Chef::Sugar::Init](https://github.com/sethvargo/chef-sugar/blob/9c8792b88042171e10cc756c76b6cc5a6124b8f7/lib/chef/sugar/init.rb) to Sugarable DSL.
That would allow us to use `systemd?`, `upstart?` and `runit?` helpers in project and software definitions. 
It's really helpful because currently, when we want to work around systemd-driven distributions, we need to use tricks like this:

```ruby
if (debian_platform? && platform_version >= '8') ||
   (ubuntu? && platform_version >= '16.04') ||
   (rhel? && platform_version >= '7')
# ...
```
After this PR it would be more clear:
```ruby
if systemd?
# ...
```

`Chef::Sugar::Init` is available in chef-sugar starting since v3.2.0, which is OK for omnibus (currently we require `chef-sugar ~>3.3`)

/cc @chef/omnibus-maintainers


legal90:
I've rebased the branch, so now it could be merged

legal90:
Rebased the branch again. Could it be merged now?

srenatus:
@legal90 thanks! 🎉

PR #775, opened 2017-05-09
Add a strip BuildCommand

atomic-penguin:
Signed-off-by: Eric G. Wolfe <eric.wolfe@gmail.com>

### Description

Add a strip build command to allow symbol stripping on binaries.  This is a cleanup build command to help produce smaller omnibus artifacts.

Example usage: `strip install_dir`

#### Maintainers

  serial number (Not sure, how would I check?)
- [x] If this change impacts compatibility with omnibus-software, the
  corresponding change is reviewed and there is a release plan (Negative)
- [x] If this change impacts compatibility with the omnibus cookbook, the
  corresponding change is reviewed and there is a release plan (Negative)


ksubrama:
Beware that there's no strip natively in windows.  Calling `command()` with `in_msys_bash` set to true (see line 189,190) or, if that doesn't work for you, following the example of how `patch()` does it (lines 247-250) is preferable.

When I get a spare moment, I'll see if I can patch this up with windows support (if you don't get to it before I do :-P).

ksubrama:
Also, it's totally reasonable to pull that extension whitelist from health_check out into a common file and add the python extensions into that.  That'll let you use it in both places.  Note to self:  do this :-P

atomic-penguin:
@ksubrama I DRY'd out the whitelist into omnibus/whitelist for use by health_check and builder.  Let me know if this is acceptable.  Also added the in_msys_bash option, as you noted.

@srenatus it turns out that find is totally faster.  Not sure if it is included on windows, inside of msys bash, however?

lib/omnibus/builder.rb
```diff
@@ -563,6 +564,29 @@ def delete(path, options = {})
...
+    # @return (see #command)
+    #
+    def strip(path)
+      regexp_ends = ".*(" + IGNORED_ENDINGS.map { |e| e.gsub(/\./, '\.') }.join("|") + ")$"
+      regexp_patterns = IGNORED_PATTERNS.map { |e| ".*" + e.gsub(/\//, '\/') + ".*" }.join("|")
+      regexp = regexp_ends + "|" + regexp_patterns
+
+      # Do not actually care if strip runs on non-strippable file, as its a no-op.  Hence the `|| true` appended.
+      # Do want to avoid stripping files unneccessarily so as not to slow down build process.
+      find_command = "find #{path}/ -type f -regextype posix-extended ! -regex '#{regexp}' | xargs strip || true"
```
ksubrama: The single quotes here will probably fail on windows because of the funky way we shellout here https://github.com/chef/omnibus/blob/master/lib/omnibus/util.rb#L92  It's a `bash -c 'your command'` that runs through cmd and they both escape stuff differently.

On the other hand, if this works on *nix systems, I'm ok with just coming back later and fighting with the windows compat issue.  We'll probably end up using your previous ruby-based algorithm on windows anyway and eating the built-time cost on windows because that's usually how this stuff works out.

ksubrama:
I'm :+1: on this assuming nobody else objects or has strong feeling about windows incompatibility.  I'm fine with dealing with that later.

lib/omnibus/builder.rb
```diff
@@ -563,6 +564,29 @@ def delete(path, options = {})
...
+    #   the path of the file(s) to strip
+    #
+    # @return (see #command)
+    #
+    def strip(path)
+      regexp_ends = ".*(" + IGNORED_ENDINGS.map { |e| e.gsub(/\./, '\.') }.join("|") + ")$"
+      regexp_patterns = IGNORED_PATTERNS.map { |e| ".*" + e.gsub(/\//, '\/') + ".*" }.join("|")
+      regexp = regexp_ends + "|" + regexp_patterns
+
+      # Do not actually care if strip runs on non-strippable file, as its a no-op.  Hence the `|| true` appended.
```
srenatus: `s/its/it's/`

atomic-penguin: From my testing, it does seem to fail on Windows.  I'm going to see if I can hack up Windows support today.

atomic-penguin: Seems like simply escaping double-quotes on the regexp string will get it this to work on Windows, without breaking it on Linux.  Just amended the commit to do that.

ksubrama: NICE!

PR #776, opened 2017-05-10
Support appbundler 0.11.0

lamont-granquist:
The changes here:

- The `lockdir` argument used to be entirely auto-populated.  It is now settable via the `lockdir` option.  The default auto-population algorithm has not changed, but it now gives a much more informative error message if that fails (I think it NoMethod on NilClass'd before).

- Support has been added for the `gem` and `without` options which are used by ChefDK and use the 3-argument version of appbundler to do transitive Gemfile.lock minimum-viable-horribleness in ChefDK.

- The API for all other non-ChefDK projects should still use the 2-argument form of appbundler and the API should be unchanged (certainly it is *intended* that it will be unchanged).

There's a bug in appbundler around trying to use the without argument with the 2-arg form, which I believe is broken in appbundler -- it'll get passed to appbundler and silently fail.  It isn't in principle nonsense that should cause a raise here, but if someone wants that fixed, it should just get fixed in appbundler.

### DIFF
--- a/lib/omnibus/builder.rb
+++ b/lib/omnibus/builder.rb
@@ -1,5 +1,5 @@
 #
-# Copyright 2012-2014 Chef Software, Inc.
+# Copyright 2012-2018, Chef Software Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


lamont-granquist:
I think this has a hard requirement on bundler 1.16.x now via a hard requirement on bundler 1.16.x in appbundler 0.11.0 which I don't see any easy way around.  If it becomes problematic we could do an appbundler 0.10.1 retro-release without the bundler 1.16.x fixes and users could manually pin onto that, but the happy path is everyone just upgrades (particularly because bundler/rubygems security vulnerabilities) and AFAIK the only projects that care about appbundling are chef and chef-dk.

thommay:
I didn't think we had a hard dep on bundler 1.16? Certainly the change i made yesterday ought to be back compat with earlier versions. In general though, we should be moving everything to 1.16 indeed so this is fine either way.

lamont-granquist:
this needs some backcompat because only chef-dk can use (and must use) the 4 arg version, while chef (and anything else that appbundles) needs the 3 arg version.

lamont-granquist:
for posterity:  i was completely wrong about the dep on bundler >= 1.16 for the 3-arg version and it all should be backcompat (even to like chef-12 era).  the 4-arg version probably does need modern bundler, but that is modern chef-dk 2/3 territory so that is fine and we ship a modern enough bundler there.

lamont-granquist:
master seems to be busted as far as i can tell, actual omnibus builds in jenkins appear to be working as much as i can make them work right now and are appbundling correctly with this.

PR #780, opened 2017-06-01
expose overrides to software defns

lamont-granquist:
allows software defns to pull other arbitrary overrides out of the software's override hash other than `:source` or `:version`


lamont-granquist:
@chef/omnibus-maintainers pretty trivial.

i didn't spec test the behavior because we already have coverage that the override hash works.

PR #787, opened 2017-06-26
Update ruby versions we test and OS we test on

tas50:
Test on the latest distros in appveyor / travis and use the latest ruby releases. Same thing we do elsewhere

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,7 @@ $RECYCLE.BIN/
 /test/version_tmp/
 /tmp/
 /spec/tmp
+gem_graph.png
 
 ## Specific to RubyMotion:
 .dat*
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -52,6 +52,7 @@ $RECYCLE.BIN/
 /test/version_tmp/
 /tmp/
 /spec/tmp
+gem_graph.png
```
tas50: I committed this file by accident earlier. We should def make sure it's ignored.

tas50:
@scotthain I force pushed this to remove the gem graph I committed by accident.

scotthain:
@tas50 looks good now - let's rebase and get it shipped.

PR #790, opened 2017-07-11
Allow setting a custom directory pattern for publishers

legal90:
### Description
This PR adds a possibility to set a custom pattern for publisher's directory. 

It's done by adding a new config option, `publish_dir_pattern`, which supports interpolation of package metadata keys, such as `platform`, `platform_version`, `arch`, `version`, `basename` and other. 
It uses a simple %-formatting, so it also supports any custom prefixes/suffixes for publisher's dir if needed.

The default values of this option are kept unchanged, so the backward compatibility is saved and unit tests should pass fine.

P.s. I'm not sure that `publish_dir_pattern` is the best option name for this case. Suggestions are welcome! ;)

/cc @chef/omnibus-maintainers

### DIFF
--- a/lib/omnibus/config.rb
+++ b/lib/omnibus/config.rb
@@ -333,13 +333,6 @@ module Omnibus
     # @return [Integer]
     default(:publish_retries, 2)
 
-    # Directory pattern for publisher. Interpolation of matadata keys are supported.
-    #
-    # @example 'custom_prefix/%{platform}/%{platform_version}/%{arch}/%{basename}'
-    #
-    # @return [String]
-    default(:publish_dir_pattern, nil)
-
     # --------------------------------------------------
     # @!endgroup
     #
@@ -377,6 +370,14 @@ module Omnibus
     default(:artifactory_base_path) do
       raise MissingRequiredAttribute.new(self, :artifactory_base_path, "'com/mycompany'")
     end
+
+    # Directory pattern for the Artifactory publisher.
+    # Interpolation of metadata keys is supported.
+    #
+    # @example '%{platform}/%{platform_version}/%{arch}/%{basename}'
+    #
+    # @return [String]
+    default(:artifactory_publish_pattern, "%{name}/%{version}/%{platform}/%{platform_version}/%{basename}")
 
     # The path on disk to an SSL pem file to sign requests with.
     #
@@ -442,6 +443,14 @@ module Omnibus
     #
     # @return [String, nil]
     default(:publish_s3_profile, nil)
+
+    # Directory pattern for the S3 publisher.
+    # Interpolation of metadata keys is supported.
+    #
+    # @example '%{platform}/%{platform_version}/%{arch}/%{basename}'
+    #
+    # @return [String]
+    default(:s3_publish_pattern, "%{platform}/%{platform_version}/%{arch}/%{basename}")
 
     # --------------------------------------------------
     # @!endgroup
--- a/lib/omnibus/config.rb
+++ b/lib/omnibus/config.rb
@@ -333,6 +333,13 @@ module Omnibus
     # @return [Integer]
     default(:publish_retries, 2)
 
+    # Directory pattern for publisher. Interpolation of matadata keys are supported.
+    #
+    # @example 'custom_prefix/%{platform}/%{platform_version}/%{arch}/%{basename}'
+    #
+    # @return [String]
+    default(:publish_dir_pattern, nil)
+
     # --------------------------------------------------
     # @!endgroup
     #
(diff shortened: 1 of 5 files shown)


lib/omnibus/config.rb
```diff
@@ -333,6 +333,13 @@ def reset!
     # @return [Integer]
     default(:publish_retries, 2)
 
+    # Directory pattern for publisher. Interpolation of matadata keys are supported.
+    #
+    # @example 'custom_prefix/%{platform}/%{platform_version}/%{arch}/%{basename}'
+    #
+    # @return [Integer]
```
srenatus: I think this should be `[String]`.

lib/omnibus/config.rb
```diff
@@ -333,6 +333,13 @@ def reset!
...
+    default(:publish_dir_pattern, nil)
```
srenatus: Would it make sense to set the default **here** instead of [having it in this line?](https://github.com/chef/omnibus/pull/790/files#diff-eee644ad5e745ef006b9318510bacb7fR271)

lib/omnibus/publishers/s3_publisher.rb
```diff
@@ -63,19 +63,20 @@ def s3_configuration
...
     #
     # @return [String]
     #
     def key_for(package, *stuff)
+      pattern = Config.publish_dir_pattern || "%{platform}/%{platform_version}/%{arch}/%{basename}"
       File.join(
-        package.metadata[:platform],
-        package.metadata[:platform_version],
-        package.metadata[:arch],
-        package.name,
```
srenatus: [Q] I'm not sure about this -- for this to work the same as before, `package.name == package.metadata[:basename]` must hold; is that the case? (It's also possible that I'm just confused.)

lib/omnibus/publishers/s3_publisher.rb
```diff
@@ -63,19 +63,20 @@ def s3_configuration
...
     # postfixed to the end of the path.
     #
+    # @example
+    #   'el/6/x86_64/chef-11.6.0-1.el6.x86_64.rpm/chef-11.6.0-1.el6.x86_64.rpm'
+    #
     # @param [Package] package
     #   the package this key is for
     # @param [Array<String>] stuff
-    #   the additional things to prepend
+    #   the additional things to append
```
srenatus: 💯

lib/omnibus/publishers/s3_publisher.rb
```diff
@@ -63,19 +63,20 @@ def s3_configuration
     # The unique upload key for this package. The additional "stuff" is
     # postfixed to the end of the path.
     #
+    # @example
```
srenatus: Thanks, adding examples here is a good idea

lib/omnibus/config.rb
srenatus: Ah, I see. I've missed that. Would it make sense to have two different settings then? `artifactory_publish_pattern` and `s3_publish_pattern`? (No strong opinion, just floating ideas.)

lib/omnibus/config.rb
```diff
@@ -333,6 +333,13 @@ def reset!
     # @return [Integer]
     default(:publish_retries, 2)
 
+    # Directory pattern for publisher. Interpolation of matadata keys are supported.
```
srenatus: [typo] `s/matadata/metadata/` 😃

lib/omnibus/config.rb
legal90: Yes, that sounds reasonable. Especially if we want to use both of publishers for the same Omnibus project. 👍 
I've updated the PR, so now we have 2 different options: `artifactory_publish_pattern` and `s3_publish_pattern`, with defaults defined in `Omnibus::Config` class

lib/omnibus/config.rb
```diff
@@ -378,6 +371,14 @@ def reset!
       raise MissingRequiredAttribute.new(self, :artifactory_base_path, "'com/mycompany'")
     end
 
+    # Directory pattern for the Artifactory publisher.
+    # Interpolation of metadata keys are supported.
```
srenatus: [nit] Should be _"Interpolation ... **is** supported"_

lib/omnibus/config.rb
```diff
@@ -443,6 +444,14 @@ def reset!
     # @return [String, nil]
     default(:publish_s3_profile, nil)
 
+    # Directory pattern for the S3 publisher.
+    # Interpolation of metadata keys are supported.
```
srenatus: [nit] same here, `s/are/is/`

legal90: Thanks! Fixed both of cases.

srenatus:
🏓 @chef/omnibus-maintainers can we get more 👀 here, please?

PR #797, opened 2017-08-15
Add config option to use path style S3 URLs

szechyjs:
### Description

Self-hosted S3 solutions like Minio don't support subdomain style bucket URLs. This adds an option to use path style S3 URLs when using S3 caching.

#### Maintainers

  serial number

### DIFF
--- a/lib/omnibus/config.rb
+++ b/lib/omnibus/config.rb
@@ -315,6 +315,11 @@ module Omnibus
     # @return [String, nil]
     default(:s3_endpoint, nil)
 
+    # Use path style URLs instead of subdomains for S3 URLs
+    #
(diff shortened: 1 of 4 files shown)


robbkidd:
tenor-223889545

Beware when using path-style on AWS S3 and not in the US-East-1 region:

> The path-style syntax, however, requires that you use the region-specific endpoint when attempting to access a bucket.
-- https://docs.aws.amazon.com/AmazonS3/latest/dev/VirtualHosting.html

Any error handling for the 307 that will be returned by AWS when attempting to access a path-style bucket when planets aren't aligned properly?

robbkidd:
KK. Just sayin' ...

bewaaaare

Issue #402, opened 2014-12-04
Add ability to sign Debian (.deb) packages.


kwilczynski:
We already sign the _RPM_ packages and the _PKG_ files when they are built by Omnibus, and so it would be nice to sign _Debian_ (`.deb`) packages too.

Hobbsee:
Useful links here:

http://blog.packagecloud.io/eng/2014/10/28/howto-gpg-sign-verify-deb-packages-apt-repositories/
https://purplefloyd.wordpress.com/2009/02/05/signing-deb-packages/

I've verified that following the instructions in these links, when using a omnibus-built deb, works correctly.

Hobbsee:
There are a couple of outstanding issues:
  - debsigs isn't installed by default.  Do we rely on the user to install this, or do we add this as a dependency of omnibus somewhere?  If the latter, how is this done?
  - debsigs doesn't seem to allow you to specify a passphrase via the command line, even with what GPG normally accepts:
```
PASSPHRASE='foobar' debsigs --sign=origin -k 803709B6 /vagrant/pkg/flapjack_1.2.1~+20141218052006-v1.2.1-trusty-1_amd64.deb
```

The tests pass.  Once the passphrase issue is sorted out, I'll test-build flapjack with this, and see how it actually works.

kwilczynski:
@Hobbsee you back look how signature is added and then do it in pure Ruby, it's trivial. This avoid dependency issues with `debsigs`, etc.

Anyway, thank you for doing this, outstanding work! :+1: :heart: :cake:

Hobbsee:
Ah, of course!  I'll take a look.

kwilczynski:
@Hobbsee you are a :star: :)

yzl:
@Hobbsee The [omnibus cookbook](https://github.com/opscode-cookbooks/omnibus) should install `debsigs`.

WarheadsSE:
Somewhat necro, I know, and apologize.
I wanted to chime in here, and let others know that GitLab has tasked me with actively adding DEB signing to Omnibus. I have work on our fork, which I will be providing as a PR here in the near future.

Thanks to all in this thread for your existing work, which I used as a basis for research and input for keeping the additions to a minimum of added cruft.

WarheadsSE:
For those of you who may be following, I have opened #799 with the changes we have implemented at GitLab, allowing us to distribute `debsigs` signed DEB packages.

kwilczynski:
[...]

PR #799, opened 2017-09-15
Add Debian (.deb) package signing (implementing debsig)

WarheadsSE:
See:

- https://gitlab.com/gitlab-org/omnibus/merge_requests/7
- https://github.com/chef/omnibus/issues/402

## What
Add Debian package signing, via methodology describe in `debsigs` documentation
https://gitlab.com/debsigs/debsigs

## How
Addition of a `sign_deb_file` function to `Packager::DEB`, after `create_deb_file`. The essential concepts of what is required to sign a `.deb` with with a `type: origin` signature is delineated per the link to `debsigs` above. There is no current functionailty built into dpkg scripting akin to `rpm --addsign`. Since the `.deb` file format is simple, we extract the contents of the archvice (`ar x`), sign the concatenated (specifically ordered) contents, and then append the created signature to the archive (`ar rc debfile _gpgorigin`).

These steps could have been accomplished in pure Ruby with the addition of several modules (GPGME, libarchive) except for two concerns: age & maintenance, `fakeroot` requirements.

Tests have been added to attempt to cover the behavior correctly.

## External Program Requirements
- `gpg` : This is already an existing requirement of `Packager::RPM` due to the use of `rpmsign`
- `ar` : Most systems that attempt to build Debian packages will have the `ar` command, and it has been confirmed that MacOS also has this utility.

Adding `ar` to the required tools compiled by Omnibus may be required.

There is *no need* to add `debsigs`/`debsig-verify` as a requirement, as we are implementing the logic in Ruby and `Shellout`.

### GPG versions
Care was taken to ensure compatibility with distribution provided binary versions of `gpg2` or `gpg` for LTS versions of distributions supported by GitLab. This list can be seen at https://gitlab.com/gitlab-org/omnibus/merge_requests/7#note_35053215 . The code is written to prefer `gpg2` if present.

### GPG Keys
GitLab experienced issues in regards to `gpg --import`, and eventually settled on `gpg --batch --no-tty --allow-secret-key-import --import` as a part of our CI job. So long as the key is present for the GPG calls (via `--homedir #{ENV['HOME']}/.gnupg`) this should not be an issue for any non-automated use. As the GPG key import process is not a part of Omnibus itself, this should be of little concern to this code changes in this MR.

Relates to https://gitlab.com/gitlab-org/omnibus-gitlab/issues/2537
Closes https://github.com/chef/omnibus/issues/402

### DIFF
--- a/docs/Building on Debian.md
+++ b/docs/Building on Debian.md
@@ -39,6 +39,7 @@ package :deb do
   license 'Apache 2.0'
   priority 'extra'
   section 'databases'
+  signing_passphrase 'acbd1234'
 end
 ```
 
@@ -46,6 +47,7 @@ Some DSL methods available include:
 
 | DSL Method           | Description                                 |
 | :------------------: | --------------------------------------------|
+| `signing_passphrase` | The passphrase to sign the RPM with         |
 | `vendor`             | The name of the package producer            |
 | `license`            | The default license for the package         |
 | `priority`           | The priority for the package                |
@@ -56,4 +58,4 @@ If you are unfamiliar with any of these terms, you should just accept the defaul
 For more information, please see the [`Packager::DEB` documentation](http://www.rubydoc.info/github/chef/omnibus/Omnibus/Packager/DEB).
 
 ### Notes on DEB-signing
-At this time, signing Debian packages is not supported.
+To sign an DEB, you will need a GPG keypair. You can [create your own signing key](http://www.madboa.com/geek/gpg-quickstart/) or [import an existing one](http://irtfweb.ifa.hawaii.edu/~lockhart/gpg/gpg-cs.html). Omnibus will automatically call `gpg` with arguments that assume the real name associated to the GPG key is the same as the name of the project maintainer as specified in your Omnibus config.
--- a/lib/omnibus/packagers/deb.rb
+++ b/lib/omnibus/packagers/deb.rb
@@ -60,11 +60,36 @@ module Omnibus
 
       # Create the deb
       create_deb_file
+
+      # Sign the deb
+      sign_deb_file
     end
 
     #
     # @!group DSL methods
     # --------------------------------------------------
+
+    #
+    # Set or return the signing passphrase. If this value is provided,
+    # Omnibus will attempt to sign the DEB.
+    #
+    # @example
+    #   signing_passphrase "foo"
+    #
+    # @param [String] val
+    #   the passphrase to use when signing the DEB
+    #
+    # @return [String]
+    #   the DEB-signing passphrase
+    #
+    def signing_passphrase(val = NULL)
+      if null?(val)
+        @signing_passphrase
+      else
+        @signing_passphrase = val
+      end
+    end
+    expose :signing_passphrase
 
     #
     # Set or return the vendor who made this package.
@@ -386,6 +411,62 @@ module Omnibus
         "-z#{compression_level} -Z#{compression_type} -S#{compression_strategy}"
       else
         "-z#{compression_level} -Z#{compression_type}"
+      end
+    end
+
+    #
+    # Sign the  +.deb+ file with gpg. This has to be done as separate steps
+    # from creating the +.deb+ file. See +debsigs+ source for behavior
+    # replicated here. +https://gitlab.com/debsigs/debsigs/blob/master/debsigs.txt#L103-124+
+    #
+    # @return [void]
+    def sign_deb_file
+      if !signing_passphrase
+        log.info(log_key) { "Signing not enabled for .deb file" }
+        return
+      end
+
+      log.info(log_key) { "Signing enabled for .deb file" }
+
+      # Check our dependencies and determine command for GnuPG. +Omnibus.which+ returns the path, or nil.
+      gpg = nil
+      if Omnibus.which("gpg2")
+        gpg = "gpg2"
+      elsif Omnibus.which("gpg")
+        gpg = "gpg"
+      end
+
+      if gpg && Omnibus.which("ar")
+        # Create a directory that will be cleaned when we leave the block
+        Dir.mktmpdir do |tmp_dir|
+          Dir.chdir(tmp_dir) do
+            # Extract the deb file contents
+            shellout!("ar x #{Config.package_dir}/#{package_name}")
+            # Concatenate contents, in order per +debsigs+ documentation.
+            shellout!("cat debian-binary control.tar.* data.tar.* > complete")
+            # Create signature (as +root+)
+            gpg_command =  "#{gpg} --armor --sign --detach-sign"
+            gpg_command << " --local-user '#{project.maintainer}'"
+            gpg_command << " --homedir #{ENV['HOME']}/.gnupg" # TODO: Make this configurable
+            ## pass the +signing_passphrase+ via +STDIN+
+            gpg_command << " --batch --no-tty"
+            ## Check `gpg` for the compatibility/need of pinentry-mode
+            # - We're calling gpg with the +--pinentry-mode+ argument, and +STDIN+ of +/dev/null+
+            # - This _will_ fail with exit code 2 no matter what. We want to check the +STDERR+
+            #   for the error message about the parameter. If it is _not present_ in the
+            #   output, then we _do_ want to add it. (If +grep -q+ is +1+, add parameter)
+            if shellout("#{gpg} --pinentry-mode loopback </dev/null 2>&1 | grep -q pinentry-mode").exitstatus == 1
+              gpg_command << " --pinentry-mode loopback"
+            end
+            gpg_command << " --passphrase-fd 0"
+            gpg_command << " -o _gpgorigin complete"
+            shellout!("fakeroot #{gpg_command}", input: signing_passphrase)
+            # Append +_gpgorigin+ to the +.deb+ file (as +root+)
+            shellout!("fakeroot ar rc #{Config.package_dir}/#{package_name} _gpgorigin")
+          end
+        end
+      else
+        log.info(log_key) { "Signing not possible. Ensure that GnuPG and GNU AR are available" }
       end
     end
 
--- a/spec/unit/packagers/deb_spec.rb
+++ b/spec/unit/packagers/deb_spec.rb
@@ -337,6 +337,46 @@ module Omnibus
       end
     end
 
+    describe "#sign_deb_file", :not_supported_on_windows do
+      context "when DEB signing is not enabled" do
+        before do
+          subject.signing_passphrase(nil)
+        end
+
+        it "logs a message" do
+          output = capture_logging { subject.sign_deb_file }
+          expect(output).to include("Signing not enabled for .deb file")
+        end
+      end
+
+      context "when DEB signing is enabled" do
+        before do
+          allow(subject).to receive(:shellout!)
+          allow(subject).to receive(:package_name).and_return("safe")
+          subject.signing_passphrase("foobar")
+        end
+
+        it "logs a message" do
+          output = capture_logging { subject.sign_deb_file }
+          expect(output).to include("Signing enabled for .deb file")
+        end
+
+        it "finds gpg and ar commands" do
+          output = capture_logging { subject.sign_deb_file }
+          expect(output).not_to include("Signing not possible.")
+        end
+
+        it "runs correct commands" do
+          expect(subject).to receive(:shellout!)
+            .at_least(:once).with(/ar x/)
+            .at_least(:once).with(/cat debian-binary control\.tar/)
+            .at_least(:once).with(/fakeroot gpg/)
+            .at_least(:once).with(/fakeroot ar rc/)
+          subject.sign_deb_file
+        end
+      end
+    end
+
     describe "#package_size" do
       before do
         project.install_dir(staging_dir)


docs/Building on Debian.md
```diff
@@ -56,4 +58,4 @@ If you are unfamiliar with any of these terms, you should just accept the defaul
 For more information, please see the [`Packager::DEB` documentation](http://www.rubydoc.info/github/chef/omnibus/Omnibus/Packager/DEB).
 
 ### Notes on DEB-signing
-At this time, signing Debian packages is not supported.
+To sign an DEB, you will need a GPG keypair. You can [create your own signing key](http://www.madboa.com/geek/gpg-quickstart/) or [import an existing one](http://irtfweb.ifa.hawaii.edu/~lockhart/gpg/gpg-cs.html). Omnibus automatically call `gpg` with arguments that assume the real name associated to the GPG key is the same as the name of the project maintainer as specified in your Omnibus config.
```
scotthain: Some grammar nitpicks here: Omnibus `will` automatically call

lib/omnibus/packagers/deb.rb
```diff
@@ -390,6 +415,73 @@ def compression_params
     end
 
     #
+    # Check for the existance of a command in +ENV['PATH']+. Exit code +127+
+    # tells us the shell could not find the program.
+    #
+    # @return [bool]
+    #
+    def check_command_exists(command)
```
scotthain: This probably belongs in the `Util` class

lib/omnibus/packagers/deb.rb
```diff
@@ -390,6 +415,73 @@ def compression_params
...
+              gpg_command << " -o _gpgorigin complete"
+              shellout!("fakeroot #{gpg_command}",
+                input: signing_passphrase
+              )
+              # Append +_gpgorigin+ to the +.deb+ file (as +root+)
+              shellout!("fakeroot ar rc #{Config.package_dir}/#{package_name} _gpgorigin")
+            end
+          }
+        else
+          log.info(log_key) { "Singing not possible. Ensure that GnuPG and GNU AR are available"}
```
scotthain: s/Singing/Signing

scotthain:
@WarheadsSE a couple other small notes: our project here uses the Developer Certificate or Origin, or DCO, to ensure that it's legally OK for us to accept your contribution. You can read more about the DCO here: https://github.com/chef/inspec/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco, but essentially every commit must have a sign-off.

It also looks like there are some style issues (you can see the specifics in the travis and appveyor) that should be pretty easy to fix up - you can also run `rake chefstyle` to verify locally.

Thanks again!

lib/omnibus/packagers/deb.rb
kwilczynski: I believe, that it actually not need to be provided at all, as there is `which` method provided and used already in various places, as per: https://github.com/chef/omnibus/blob/master/lib/omnibus.rb#L170; and we should see whether it could be used here instead of `chek_command_exists`.

lib/omnibus/packagers/deb.rb
```diff
@@ -390,6 +415,73 @@ def compression_params
...
+    end
+
+    #
+    # Sign the  +.deb+ file with gpg. This has to be done as separate steps
+    # from creating the +.deb+ file. See +debsigs+ source for behavior
+    # replicated here. +https://gitlab.com/debsigs/debsigs/blob/master/debsigs.txt#L103-124+
+    #
+    # @return [void]
+    def sign_deb_file
+      if signing_passphrase
```
kwilczynski: You could reduce the level of nesting by checking if signing passphrase is provided first and then bail out early (an "early return"). It would improve readability a bit.

lib/omnibus/packagers/deb.rb
```diff
@@ -390,6 +415,73 @@ def compression_params
...
+        gpg = nil
+        if check_command_exists('gpg2')
+          gpg = 'gpg2'
+        elsif check_command_exists('gpg')
+          gpg = 'gpg'
+        end
+
+        if gpg && check_command_exists('ar')
+          # Create a directory that will be cleaned when we leave the block
+          Dir.mktmpdir { |tmp_dir|
```
kwilczynski: I would make it a `do ... end` block, since there is a lot of code inside this one, and curly bracket can get lost in translation there so to speak.

lib/omnibus/packagers/deb.rb
```diff
@@ -390,6 +415,73 @@ def compression_params
...
+              if shellout("#{gpg} --pinentry-mode loopback </dev/null 2>&1 | grep -q pinentry-mode").exitstatus == 1
+                gpg_command << " --pinentry-mode loopback"
+              end
+              gpg_command << " --passphrase-fd 0"
+              gpg_command << " -o _gpgorigin complete"
+              shellout!("fakeroot #{gpg_command}",
+                input: signing_passphrase
+              )
+              # Append +_gpgorigin+ to the +.deb+ file (as +root+)
+              shellout!("fakeroot ar rc #{Config.package_dir}/#{package_name} _gpgorigin")
```
kwilczynski: The `fakeroot` binary might not be installed on some systems, thus to be overly cautious you could also check for its existence - similarly to `gpg`/`gpg2` and `ar`.

lib/omnibus/packagers/deb.rb
kwilczynski: Space at the end before closing bracket.

lib/omnibus/packagers/deb.rb
```diff
@@ -390,6 +415,73 @@ def compression_params
...
+            end
+          }
+        else
+          log.info(log_key) { "Singing not possible. Ensure that GnuPG and GNU AR are available"}
+        end
+      else
+        log.info(log_key) { "Signing not enabled for .deb file"}
```
kwilczynski: Space at the end before closing bracket.

lib/omnibus/packagers/deb.rb
```diff
@@ -390,6 +415,73 @@ def compression_params
...
+          Dir.mktmpdir { |tmp_dir|
+            Dir.chdir(tmp_dir) do
+              # Extract the deb file contents
+              shellout!("ar x #{Config.package_dir}/#{package_name}")
+              # Concantenate contents, in order per +debsigs+ documentation.
+              shellout!("cat debian-binary control.tar.* data.tar.* > complete")
+              # Create signature (as +root+)
+              gpg_command =  "#{gpg} --armor --sign --detach-sign"
+              gpg_command << " --local-user '#{project.maintainer}'"
+              gpg_command << " --homedir #{ENV['HOME']}/.gnupg" # TODO: Make this configurable
```
kwilczynski: Agreed, it should be a configuration option, or at least an environment variable.

lib/omnibus/packagers/deb.rb
```diff
@@ -390,6 +415,73 @@ def compression_params
...
+              ## pass the +signing_passphrase+ via +STDIN+
+              gpg_command << " --batch --no-tty"
+              ## Check `gpg` for the compatibility/need of pinentry-mode
+              # - We're calling gpg with the +--pinentry-mode+ argument, and +STDIN+ of +/dev/null+
+              # - This _will_ fail with exit code 2 no matter what. We want to check the +STDERR+
+              #   for the error message about the parameter. If it is _not present_ in the
+              #   output, then we _do_ want to add it. (If +grep -q+ is +1+, add parameter)
+              if shellout("#{gpg} --pinentry-mode loopback </dev/null 2>&1 | grep -q pinentry-mode").exitstatus == 1
```
kwilczynski: You could parse output from `gpg` using Ruby (in Ruby?) over using `grep`, but this is more of a nit pick.

docs/Building on Debian.md
WarheadsSE: Fixed.

kwilczynski:
@WarheadsSE looks good, with some comments here and there.

I wanted to add one thing, since you are using external `ar` binary to compress files. Have a look at this (quite old now) thread as not every `ar` is created equal, so to speak, which is something possible worth to note somewhere (or not, it's up to you): https://bugs.launchpad.net/ubuntu/+source/dpkg-sig/+bug/1156988

lib/omnibus/packagers/deb.rb
WarheadsSE: I'd agree with @kwilczynski here. I can likely refactor the reliant code to make use of the existing `which`, as this essentially performs the same tasks, but only returns boolean vs a string/nil.

lib/omnibus/packagers/deb.rb
WarheadsSE: Done. bad/old habit.

lib/omnibus/packagers/deb.rb
scotthain: we're actually guaranteed that `fakeroot` is present in the omnibus cookbook here: https://github.com/chef-cookbooks/omnibus/blob/master/recipes/_packaging.rb#L32

WarheadsSE: Indeed this caused test failures above. I'll have to recheck my logic on why I did this with `fakeroot`.

lib/omnibus/packagers/deb.rb
WarheadsSE: While I agree I could pull the output into Ruby, parse it, et al. It seems much simpler to get everything in one swoop, as opposed to complicating the state of the code.

lib/omnibus/packagers/deb.rb
WarheadsSE: As of now, it lines up with the RPM packager: lib/omnibus/packagers/rpm.rb:438

lib/omnibus/packagers/deb.rb
WarheadsSE: It appears I did this so that the archive (deb file) behaves as expected. Considering best practices for deb files to have their contents owned by root. It is also used by the actual `create_deb_file` function that is already a part of the package.

lib/omnibus/packagers/deb.rb
kwilczynski: Oh I see, OK.

lib/omnibus/packagers/deb.rb
kwilczynski: @scotthain ah, OK. Albeit, not everyone would use the cookbook. I personally don't use it in our internal build system (at work at the moment). Which is where the idea came from.

Nonetheless, should there be a parallel Pull Request against the cookbook to install GnuPG and `ar`? What do you think, @scotthain and @WarheadsSE?

lib/omnibus/packagers/deb.rb
kwilczynski: No worries!

lib/omnibus/packagers/deb.rb
scotthain: @kwilczynski Yeah a PR in the omnibus cookbook would be fantastic, just for safety's sake. Currently they are part of the base install of the systems we use internally at Chef but that could easily change in the future.

kwilczynski: @scotthain Done. Pull Request opened at https://github.com/chef-cookbooks/omnibus/pull/217.

WarheadsSE:
I am hunting a misbehaving test at the moment. Somehow, it is not properly receiving the final `shellout` to `fakeroot ar rc`, and I am uncertain as to how that might be the case.

WarheadsSE:
The `ar` 'problem' mentioned by @kwilczynski is solved by his PR to add it to the cookbooks: `https://github.com/chef-cookbooks/omnibus/pull/217`

lib/omnibus/packagers/deb.rb
WarheadsSE: :+1:

WarheadsSE:
Seem my test are failing currently due to the lack of either GnuPG or `ar` in the test environment. This may remain pending until the PR from @kwilczynski  is present.

WarheadsSE:
@rhass I'm just confirming that the DCO is cleared over here, and I will get this sorted 👍

WarheadsSE:
Currently, TravisCI is failing with issues about build flags for SLES, and AppVeyor is failing due to the missing items. https://github.com/chef-cookbooks/omnibus/pull/217 has been merged, so I'd assume I need to update something here to get that sorted? @kwilczynski @rhass ?

kwilczynski:
@WarheadsSE the SLES failures need to be fixed, definitely, but it's not affecting your unit test.

Now, the AppVeyor is running the unit test on Windows, which clearly does not have neither `gnupg`, `fakeroot` nor `ar` installed by default (unless they can be installed from a third-part if a binary compiled for Win32 is available, perhaps https://chocolatey.org will have some, I am not sure, though.

If you really want to tackle this, then you would have to make GPG signing either not available on Windows, or fix the whole process to make it cross-platform (does it even make sense? who builds RPMs and DEBs on Windows?!), or skip the unit test from running on Windows.

I wonder how RPM packages handles this? Does it support/work on Windows?

WarheadsSE:
Ah, Windows. I won't be able to get away from the need for `fakeroot`, nor `ar`. It does look like the existing code in `deb_spec.rb` has operations set to not run on Windows via `:not_supported_on_windows` where it checking exact file permissions.

WarheadsSE:
Any further comments @kwilczynski @rhass ?

WarheadsSE:
Anything from your side @scotthain ?

rhass:
@WarheadsSE Unfortunately, Scott is on vacation until 10/6.

@chef/omnibus-maintainers Can I get another reviewer approval on this so we can merge this into master? Thanks!

rhass:
@WarheadsSE 

Yay

Thank you for your awesome contribution!

PR #804, opened 2017-11-09
with_clean_env should clean BUNDLER_ variables

stevendanna:
In version 1.13.0, bundler moved from BUNDLE_ to BUNDLER_ as the
prefix for their environment variables.

#### Maintainers

  serial number


stevendanna:
This appears to fix recent chef-server build failures and possibly chefdk 1 build failures.  I'm still sorting out why exactly this isn't broken in more places.

Issue #481, opened 2015-05-14
Update Omnibus::BuildVersion#build_start_time to use BUILD_TIMESTAMP


schisamo:
Jenkins 1.597+ changed the format of `BUILD_ID` from being a timestamp to the same value as `BUILD_NUMBER` which is an Integer. Omnibus looks for `BUILD_ID` but expects it to be in the old format of `yyyy-MM-dd_HH-mm-ss`. 

Chef Software Inc.'s workaround is to use the [Zen Timestamp plugin](https://wiki.jenkins-ci.org/display/JENKINS/ZenTimestamp+Plugin) to generate a new `BUILD_TIMESTAMP` value in the expected timestamp format. Omnibus will also be updated to look for `BUILD_TIMESTAMP` but we’ll also set `BUILD_ID` to the value of `BUILD_TIMESTAMP` for projects stuck on older versions of Omnibus.

See [JENKINS-26520](https://issues.jenkins-ci.org/browse/JENKINS-26520) for more details on the `BUILD_ID` change and it’s fallout!

schisamo:
We'll need the code to be smart enough to prefer `BUILD_TIMESTAMP` and fall back to `BUILD_ID`. I still think it's a good idea for it to throw an exception if parsing of either one fails.

lamont-granquist:
looks fixed

PR #805, opened 2017-11-17
Use BUILD_TIMESTAMP instead of BUILD_ID

aojea:
As explained in #481 Jenkins has changed the format of BUILD_ID
from being a timestamp to the same value as BUILD_NUMBER.

This patch uses BUILD_TIMESTAMP instead of BUILD_ID but don't use
any fall back.

Signed-off-by: itsuugo <antonio.ojea.garcia@gmail.com>

### DIFF
--- a/lib/omnibus/build_version.rb
+++ b/lib/omnibus/build_version.rb
@@ -125,8 +125,8 @@ module Omnibus
     end
 
     # We'll attempt to retrieve the timestamp from the Jenkin's set BUILD_TIMESTAMP
-    # environment variable. This will ensure platform specfic packages for the
-    # same build will share the same timestamp.
+    # or fall back to BUILD_ID environment variable. This will ensure platform specfic
+    # packages for the same build will share the same timestamp.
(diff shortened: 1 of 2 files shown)


lib/omnibus/build_version.rb
```diff
@@ -124,16 +124,16 @@ def semver
       build_tag
     end
 
-    # We'll attempt to retrive the timestamp from the Jenkin's set BUILD_ID
+    # We'll attempt to retrive the timestamp from the Jenkin's set BUILD_TIMESTAMP
```
srenatus: [nit] since you're touching that line, can you make it say "retrieve", please? 😃

srenatus:
@itsuugo also: thanks for your contribution 🎉

lib/omnibus/build_version.rb
itsuugo: Sure

itsuugo:
@scotthain I added a fallback to the old behavior

PR #811, opened 2018-01-23
Enable expeditor for builds and versioning

scotthain:
Signed-off-by: Scott Hain <shain@chef.io>

### Description

#### Maintainers

  serial number

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -4,13 +4,21 @@ slack:
 github:
   maintainer_group: chef/omnibus-maintainers
   version_tag_format: "v{{version}}"
+  minor_bump_labels:
+    - "Version: Bump Minor"
 
 merge_actions:
   - built_in:bump_version:
       ignore_labels:
         - "Version: Skip Bump"
+        - "Expeditor: Skip All"
   - bash:.expeditor/update_version.sh:
       only_if:
         - built_in:bump_version
-  - built_in:build_gem
-
+  - built_in:update_changelog:
+      ignore_labels:
+        - "Changelog: Skip Update"
+        - "Expeditor: Skip All"
+  - built_in:build_gem:
+      only_if:
+        - built_in:bump_version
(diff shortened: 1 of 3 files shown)


scotthain:
@tduffield

.expeditor/config.yml
```diff
@@ -0,0 +1,16 @@
...
+  notify_channel: omnibus-rfr
+
+github:
+  maintainer_group: chef/omnibus-maintainers
+  version_tag_format: "v{{version}}"
+
+merge_actions:
+  - built_in:bump_version:
+      ignore_labels:
+        - "Version: Skip Bump"
```
schisamo: Idiomatically we usually add a `Expeditor: Skip All` under each merge action. I wonder if we should consider adding this as built-in behavior @tduffield ?

.expeditor/config.yml
```diff
@@ -0,0 +1,16 @@
+slack:
+  notify_channel: omnibus-rfr
+
+github:
+  maintainer_group: chef/omnibus-maintainers
+  version_tag_format: "v{{version}}"
```
schisamo: We should add the following under here:

```YAML
minor_bump_labels:
    - "Version: Bump Minor"
```

scotthain:
@tduffield @schisamo I think I've incorporated your feedback - thanks!

.expeditor/config.yml
tduffield: @schisamo I go back and forth on that. My initial design intention was to be very explicit in this YAML about what labels do what.

.expeditor/config.yml
```diff
@@ -4,13 +4,24 @@ slack:
...
         - built_in:bump_version
-  - built_in:build_gem
+  - built_in:update_changelog:
+      ignore_labels:
+        - "Changelog: Skip Update"
+        - "Expeditor: Skip All"
+  - built_in:build_gem:
+      only_if:
+        - built_in:bump_version
+      ignore_labels:
```
jaymalasinha: The ignore_labels here would not take effect due to the ```only_if``` right.  Do we still need it here ?

scotthain: You're right, we don't.

PR #814, opened 2018-01-26
sometimes we just need to gorram make a link

thommay:


lamont-granquist:
is the check actually ever really needed?  could we just warn and then always create the link instead?  i've hit this before and wondered what the point of it was...

thommay:
> is the check actually ever really needed? could we just warn and then always create the link instead? i've hit this before and wondered what the point of it was...

I think it's just cut'n'waste from copy, above. But it was a bigger change than i really wanted to eat.

PR #816, opened 2018-01-26
Add config to allow rubygems publish and promote

scotthain:
Signed-off-by: Scott Hain <shain@chef.io>

### Description
@chef/jex-team 

#### Maintainers

  serial number

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -7,6 +7,9 @@ github:
   minor_bump_labels:
     - "Version: Bump Minor"
 
+rubygems:
+  - omnibus
+
 merge_actions:
   - built_in:bump_version:
       ignore_labels:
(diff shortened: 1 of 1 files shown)


.expeditor/config.yml
```diff
@@ -22,3 +25,7 @@ merge_actions:
   - built_in:build_gem:
       only_if:
         - built_in:bump_version
+
+promote:
+  action:
```
schisamo: Shouldn't this be:

```yaml
artifact_actions:
  promoted_to_stable:
    - built_in:publish_rubygems
```

That matches what we have in various other projects:
https://github.com/chef/chef/blob/master/.expeditor/config.yml#L71-L73
https://github.com/chef/chef-dk/blob/master/.expeditor/config.yml#L66-L68
https://github.com/chef/inspec/blob/master/.expeditor/config.yml#L59-L60

scotthain: I talked to @tduffield about this - since we have no artifacts (rpms, etc) there is no 'artifact_action' for this, thus we need the `promote` primitive.

schisamo: Sounds good.

PR #820, opened 2018-02-08
Avoid using git to determine gem contents

tduffield:
Signed-off-by: Tom Duffield <tom@chef.io>

### Description

Avoid using git to determine file contents because it can pick up unexpected files during CI builds.

#### Maintainers

  serial number


lamont-granquist:
👍

PR #828, opened 2018-03-13
SPIKE: Record the hardlinks in the cache commit message, so we can restore them

richardc:
git doesn't care about hardlinks; when it does the checkout it creates different files.  This isn't ideal as we end up with duplicates.

### Description

#### Maintainers

  serial number


richardc:
I think this is now in a solid enough state to be considered.  Thank you maintainers.

PR #834, opened 2018-04-13
Change exclude to use `**` to skip levels

### LINKED ISSUE #833: Exclude overmatches directory globs
### Description

Exclude seems to overmatch directory levels

### Omnibus Version

v5.6.10

### Platform Version

CentOS 7.  x86_64

### Replication Case

Adding the following to a project that installs many rubygems:

```ruby
exclude 'embedded/lib/ruby/gems/*/cache'
```

Excluded not only the intended `embedded/lib/ruby/gems/2.3.0/cache` directory, but also `embedded/lib/ruby/gems/2.3.0/gems/sprockets-3.7.1/lib/sprockets/cache` directory, which was alarming and broke the resulting package.

My expectation would be that I'd need to specify `**` to get that behaviour:

```ruby
exclude 'embedded/lib/ruby/gems/**/cache'
```

Looking at the code I think that if [the call to `File.fnmatch?()` in the exclusion matching code](https://github.com/chef/omnibus/blob/7f3652f85986ba24d681493bf2968326bfc2de46/lib/omnibus/file_syncer.rb#L65) specified additionally `File::FNM_PATHNAME` then I'd get the behaviour I expect.

Is my expectation fair?  If so I'm happy to provide a patch.

richardc:
### Description

As raised in #833 I find the current behaviour of exclude to be a little alarming, I expect the exclude `'start/*/end'` to match `'start/middle/end'` but not `'start/on/the/way/to/the/end'`

Here we specify File::FNM_PATHNAME to the fnmatch call, so `'start/*/end'` still matches `'start/middle/end'`, but you need `'start/**/end'` if you want to match  `'start/on/the/way/to/the/end'`

Closes #833 

#### Maintainers

  serial number

### DIFF
--- a/lib/omnibus/file_syncer.rb
+++ b/lib/omnibus/file_syncer.rb
@@ -56,13 +56,13 @@ module Omnibus
     #
     def all_files_under(source, options = {})
       excludes = Array(options[:exclude]).map do |exclude|
-        [exclude, "#{exclude}/*"]
+        [exclude, "#{exclude}/**"]
       end.flatten
 
       source_files = glob(File.join(source, "**/*"))
       source_files = source_files.reject do |source_file|
         basename = relative_path_for(source_file, source)
-        excludes.any? { |exclude| File.fnmatch?(exclude, basename, File::FNM_DOTMATCH) }
+        excludes.any? { |exclude| File.fnmatch?(exclude, basename, File::FNM_DOTMATCH | File::FNM_PATHNAME) }
       end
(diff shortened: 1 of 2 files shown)


twk3:
This is still an issue that is biting us. What further needs to be done for this PR?

dustinmm80:
Looks like CI may have stalled out.

tas50:
Thanks @richardc

Issue #836, opened 2018-04-30
Net fetcher: Collision of locally cached file names


legal90:
### Description

If there are software definitions having the same base names of `source[:url]`, then the fetching phase the one will override all others.

### Omnibus Version

v5.6.10

### Platform Version

Ubuntu 16.04 (but that doesn't actually matter)

### Replication Case

I have 2 software definitions with slightly different values of `source[:url]`:
  1) Software `chromedriver-228`: `https://chromedriver.storage.googleapis.com/2.28/chromedriver_linux64.zip`
  2) Software `chromedriver-237`: `https://chromedriver.storage.googleapis.com/2.37/chromedriver_linux64.zip`

I want both of them to be present in my project. However, due to the current implementation of the `NetFetcher`, both of these files will be saved to the same path: `/var/cache/omnibus/cache/chromedriver_linux64.zip`. In the result the build will be successful, but both softwares will be unpacked from the same archive, which is completely wrong.

That happens because the net fetcher takes the filename from the URL, expecting that it is unique within the whole project:
https://github.com/chef/omnibus/blob/5804316ce2613cd37a12e58e752bcfdcb48f00a0/lib/omnibus/fetchers/net_fetcher.rb#L120-L123

-------------------------------------

Should we add a new option for `source` hash, so it will be possible to define the filename for a local cache?

PR #837, opened 2018-05-02
Add "cached_name" option for net fetcher

legal90:
Fixes https://github.com/chef/omnibus/issues/836

I've added `cached_name` option for `Software#source` allowing to specify a custom filename for the source file when it is stored in the local cache. So that it could be used to work aroung a name collision issue described in #836


legal90:
Is anyone interested in this fix?

tas50:
Ping @scotthain

legal90:
@scotthain Sure! Done :)

PR #842, opened 2018-07-12
add support for light to delay validation

jmartin-r7:
### Description

Add support for delayed validation in WiX `light.exe` tool, see `-sval` option.
> Build the MSI that will be run against Smoke. Pass the -sval argument to delay validation until Smoke is run.

usage:
```
package :msi do
  wix_light_delay_validation true
end
```

--------------------------------------------------

#### Maintainers

Please ensure that you check for:

- [ ] If this change impacts git cache validity, it bumps the git cache
  serial number
- [ ] If this change impacts compatibility with omnibus-software, the
  corresponding change is reviewed and there is a release plan
- [ ] If this change impacts compatibility with the omnibus cookbook, the
  corresponding change is reviewed and there is a release plan

PR #843, opened 2018-07-12
allow project configuration to skip packagers

jmartin-r7:
### Description

Since windows builds now offer multiple packagers to execute, allow the project configuration to `skip` a packager if the maintainer does not want to build that package type.

In original APPX support PR #675, there was mention of discussion on how to supporting multiple package type for a single platform being addressed during rollout, since I don't see another existing option, this is my proposal on how to add user input to determine which package types are desired.

usage in a project DSL to omit APPX build:
```
package :appx do
  skip_packager true
end
```

#### Maintainers

  serial number

### DIFF
--- a/lib/omnibus/packagers/base.rb
+++ b/lib/omnibus/packagers/base.rb
@@ -128,6 +128,25 @@ module Omnibus
     expose :windows_safe_path
 
     #
+    # Skip this packager during build process
+    #
+    # @example
+    #   skip_package true
+    #
+    # @param [TrueClass, FalseClass] value
+    #   whether to delay validation or not
+    #
+    # @return [TrueClass, FalseClass]
+    #   whether to skip this packager type or not
+    def skip_packager(val = false)
+      unless val.is_a?(TrueClass) || val.is_a?(FalseClass)
+        raise InvalidValue.new(:iwix_light_delay_validation, "be TrueClass or FalseClass")
+      end
+      @skip_package ||= val
+    end
+    expose :skip_packager
+
+    #
     # @!endgroup
     # --------------------------------------------------
 
--- a/lib/omnibus/packagers/base.rb
+++ b/lib/omnibus/packagers/base.rb
@@ -140,7 +140,7 @@ module Omnibus
     #   whether to skip this packager type or not
     def skip_packager(val = false)
       unless val.is_a?(TrueClass) || val.is_a?(FalseClass)
-        raise InvalidValue.new(:iwix_light_delay_validation, "be TrueClass or FalseClass")
+        raise InvalidValue.new(:skip_packager, "be TrueClass or FalseClass")
       end
       @skip_package ||= val
     end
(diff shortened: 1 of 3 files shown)


lib/omnibus/packagers/base.rb
```diff
@@ -127,6 +127,25 @@ def install_dir
...
+    #   skip_package true
+    #
+    # @param [TrueClass, FalseClass] value
+    #   whether to delay validation or not
+    #
+    # @return [TrueClass, FalseClass]
+    #   whether to skip this packager type or not
+    def skip_packager(val = false)
+      unless val.is_a?(TrueClass) || val.is_a?(FalseClass)
+        raise InvalidValue.new(:iwix_light_delay_validation, "be TrueClass or FalseClass")
```
robbkidd: The first parameter to `InvalidValue.new` should be [a symbol of the method name](https://github.com/chef/omnibus/blob/3eefb1cd8de69b1d97de2962f779512892e9296d/lib/omnibus/exceptions.rb#L158-L159) that received the invalid value. So `:iwix_light_delay_validation` should probably be `:skip_packager` here.

lib/omnibus/packagers/base.rb
```diff
@@ -127,6 +127,25 @@ def install_dir
...
+    #
+    # @example
+    #   skip_package true
+    #
+    # @param [TrueClass, FalseClass] value
+    #   whether to delay validation or not
+    #
+    # @return [TrueClass, FalseClass]
+    #   whether to skip this packager type or not
+    def skip_packager(val = false)
```
tyler-ball: I would propose we change the signature of this method to be `def skip_packager()`. IMO if the user is adding `skip_packager` to their packaging definition that is enough indication of what they want to do. The `true` or `false` does not seem necessary

jmartin-r7: I like the idea, however there are other options exposed on `msi` packagers that I based this on. 

```
git grep "def.*\(val = false\)"
lib/omnibus/packagers/msi.rb:    def bundle_msi(val = false)
lib/omnibus/packagers/msi.rb:    def fast_msi(val = false)
```

I believe consistency is key to adoption of function.  I offered another similar option in #842 and `master` currently offers the above two instances.

If a refactor is desired can I consolidate my 2 PRs and expand to refactor the other two instances which would make this a breaking api change, *or* this and #843 can land as is, then offer a refactor PR that includes the required version bump.

jmartin-r7:
This seems to have gotten a number of approvals, can someone take a look here and at #842 to see if we can get these merged?

marcparadise:
Sorry this has taken so long, thanks for being patient with us.  I'm working on getting this merged today.

marcparadise:
@jmartin-r7 would you be able to rebase this on master? If I do it locally, I'll have to make a new branch and replace the PR.

jmartin-r7:
I can rebase, there are no conflicts upstream as yet, will hold an hour or so encase would rather avoid need to get approvals again.

marcparadise:
@jmartin-r7  https://github.com/chef/omnibus/pull/875 will be merging shortly, then it will be safe to rebase. It  fixes  the CI errors you'll otherwise get after a rebase on current master.

PR #858, opened 2018-10-09
Unpin gcc-4.8 for later versions of Suse

jaymalasinha:

### Description

Suse platform versions later than 12 should compile using the default gcc version.
Only use gcc-4.8 for Suse platforms before version 12.

#### Maintainers

  serial number

### DIFF
--- a/lib/omnibus/software.rb
+++ b/lib/omnibus/software.rb
@@ -700,7 +700,7 @@ module Omnibus
             "CFLAGS" => "-I#{install_dir}/embedded/include -O2",
           }
           # Enable gcc version 4.8 if it is available
-          if which("gcc-4.8")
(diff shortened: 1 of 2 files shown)


lib/omnibus/software.rb
```diff
@@ -700,7 +700,7 @@ def with_standard_compiler_flags(env = {}, opts = {})
             "CFLAGS" => "-I#{install_dir}/embedded/include -O2",
           }
           # Enable gcc version 4.8 if it is available
-          if which("gcc-4.8")
+          if which("gcc-4.8") && platform_version.satisfies?("<= 11.4")
```
scotthain: Let's make this < 12, just in case there's ever an 11.5

jaymalasinha:
@scotthain updated as per your review comment.

PR #864, opened 2018-12-07
Fix comparisons for semver style git tags

lamont-granquist:
This fixes this issue:

```
  [Software: ohai] W | 2018-12-07T12:41:02-05:00 | Version v14.8.10 for software ohai was not parseable. Comparison methods such as #satisfies? will not be available for this version.
```

### DIFF
--- a/lib/omnibus/software.rb
+++ b/lib/omnibus/software.rb
@@ -1,5 +1,5 @@
 #
-# Copyright 2012-2017, Chef Software Inc.
+# Copyright 2012-2018, Chef Software Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


lamont-granquist:
note that this supports "v10" and "v10.11" in addition to "v10.11.12", the former of which are not technically semver -- so the scope of this PR is technically broader than its title -- but since "v<anything>" is currently not parseable at all if stripping the leading v off of it results in something that is still not parseable it didn't start off parseable anyway so we're no worse off by being less strict.

lamont-granquist:
Note that this is being done tactically and should probably be replaced by mixlib-versioning, but we should probably go through our four different version comparison libraries and combine the features and kill off two of the remaining three (semverse is already retired):

* https://github.com/chef/mixlib-versioning
* https://github.com/berkshelf/semverse
* https://github.com/sethvargo/chef-sugar/blob/master/lib/chef/sugar/constraints.rb
* https://github.com/chef/chef/blob/master/lib/chef/version_class.rb

I also wouldn't be too shocked if the ruby ecosystem had a better version comparison library floating around that had been produced sometime in the past 10 years...

PR #865, opened 2018-12-10
Revert "Fix comparisons for semver style git tags"

lamont-granquist:
So that didn't work.  Those versions get passed into the git fetcher with the result:

```
   [GitFetcher] I | 2018-12-10T15:48:43+00:00 | $ git ls-remote "https://github.com/chef/chef.git" "13.11.3*"
Could not resolve `13.11.3' to a valid git SHA-1.
```


lamont-granquist:
as a frequent reviewer, so do i...

PR #878, opened 2019-02-25
Remove custom LDFLAGS and CFLAGS for osx.

jonsmorrow:
### Description

OSX 10.5 or greater can use '-Wl,-rpath,/path/to/lib' so we can use
the default flags.

Signed-off-by: Jon Morrow <jmorrow@chef.io>

#### Maintainers

  serial number


lamont-granquist:
got relevant busted specs.

lamont-granquist:
(you're good to go if you just fix them though)

PR #879, opened 2019-03-18
Loosen mixlib-shellout/ohai deps and upgrade chefstyle

tas50:
Loosen pins to allow for the latest deps.


.rubocop.yml
```diff
@@ -0,0 +1,852 @@
...
+Bundler/OrderedGems:
+  Exclude:
+    - 'Gemfile'
+
+# Cop supports --auto-correct.
+# Configuration parameters: TreatCommentsAsGroupSeparators, Include.
+# Include: **/*.gemspec
+Gemspec/OrderedDependencies:
+  Exclude:
+    - 'omnibus.gemspec'
```
lamont-granquist: we don't ever want to enable these ones...

tas50: I fixed that whole file up. I think I ran it with rubocop by accident

lamont-granquist:
some sort of license_scout meltdown only on ruby 2.4

lamont-granquist:
also on 2.5 on appveyor though

tas50:
the new ~> heredoc is not behaving as expected there on Ruby 2.4. Really odd

tas50:
Merging based on Slack approval

PR #885, opened 2019-04-25
Switch off my fork of rspec-json_expectations & fix spec failures

tas50:
The upstream released the fix we needed.

This also fixes spec failures to get the build green. It forces the spec encoding to UTF 8 to prevent failures in the new Buildkite containers, adds a missing array value that was causing a spec failure, and configures rspec to give us longer diffs when a comparison fails.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,3 +1,5 @@
+Encoding.default_external = Encoding::UTF_8
+
 require "rspec"
 require "rspec/its"
 require "rspec/json_expectations"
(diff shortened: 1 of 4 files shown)


spec/spec_helper.rb
```diff
@@ -1,3 +1,5 @@
+Encoding.default_external = Encoding::UTF_8
```
lamont-granquist: you should probably put this in the code though:

https://github.com/chef/omnibus/blob/master/lib/omnibus.rb

lamont-granquist: https://github.com/berkshelf/berkshelf/blob/master/lib/berkshelf.rb#L28-L29

tas50: I'm open to considering that for our greater set of projects.

PR #889, opened 2019-05-09
CentOS packages for Raspberry Pi 3s are 'armv7hl' not 'armv7l'

mattray:
Signed-off-by: Matt Ray <matthewhray@gmail.com>

### Description

CentOS provides packages for the Raspberry Pi 3 platform, but the arch is 'armv7hl' not 'armv7l'. If other 32-bit ARM platforms using RPMs need to be supported in the future, we can update as necessary.
https://wiki.centos.org/SpecialInterestGroup/AltArch/Arm32/RaspberryPi3

#### Maintainers

  serial number

### DIFF
--- a/lib/omnibus/packagers/rpm.rb
+++ b/lib/omnibus/packagers/rpm.rb
@@ -630,6 +630,8 @@ module Omnibus
       case Ohai["kernel"]["machine"]
       when "i686"
         "i386"
+      when "armv7l" # raspberry pi 3 CentOS
+        "armv7hl"
       when "armv6l"
         if Ohai["platform"] == "pidora"
           "armv6hl"


lib/omnibus/packagers/rpm.rb
```diff
@@ -630,6 +630,8 @@ def safe_architecture
       case Ohai["kernel"]["machine"]
       when "i686"
         "i386"
+      when "armv7l" # raspberry pi 3 CentOS
+        "armv7hl"
       when "armv6l"
         if Ohai["platform"] == "pidora"
```
lamont-granquist: do you need to replicate the same logic as down here?

mattray: I don't have 'pidora', so I figured I'd leave it on a simple path until someone shows up with another 32-bit ARM RPM-based distro that needs support. I'm testing with CentOS 7.6.

mattray: @lamont-granquist do you consider this a blocker, or can we merge it?

PR #892, opened 2019-06-15
Update the rpm package signing to work with newer rpm version

twk3:
### Description

Newer versions of rpmsign (rpm 4.13+), no longer prompt for the pass
phrase, but instead provide gpg with access to stdin/out to request it.

See: https://github.com/rpm-software-management/rpm/commit/0bce5fcf270711a2e077fba0fb7c5979ea007eb5#diff-c9c5aab99001a72b592ce7465ebff8a6

Because this process is deferred, you will also often end up with
warning output from rpmbuild before the prompt, so we can no longer
assume the prompt will be the first output.

This issue was discovered while trying to build rpm packages for
OpenSUSE 15 which has the newer rpm.

#### Maintainers

  serial number

### DIFF
--- a/resources/rpm/signing.erb
+++ b/resources/rpm/signing.erb
@@ -11,25 +11,14 @@ require 'pty'
 
 puts rpm_cmd
 PTY.spawn(rpm_cmd) do |r, w, pid|
-  prompt = r.read(19)
-
-  # match the expected prompt exactly, since that's the only way we know if
-  # something went wrong.
-  unless prompt == 'Enter pass phrase: '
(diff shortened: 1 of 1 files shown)


twk3:
This isn't ready yet. Works for the new rpmsign, but not the old. Probably because the older one does the prompt without a newline. Will test that now.

twk3:
Updated, confirmed this is working for centos 6, centos 7, opensuse 42.3, and opensuse 15

tas50:
@twk3 Thanks for the detailed explanation and confirming this doesn't break existing systems.

PR #895, opened 2019-07-24
Separate removing writable dmg

jaymalasinha:
Taking three step approach to solving this:
* Separate the remove writeable dmg call
* puppetstrings option enables progress output on hdiutil
* Verify the checksum for the final image created
* check detach completes before compressing the dmg

These put together shouldresolve the hdiutil resource busy failure
during compress_dmg

### Description

#### Maintainers

  serial number

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -4,7 +4,6 @@
 slack:
   notify_channel:
     - chef-found-notify
-    - omnibus-rfr
 
 # This publish is triggered by the `built_in:publish_rubygems` artifact_action.
 rubygems:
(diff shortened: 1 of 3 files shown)


btm:
I'm concerned this might not fix the current issue because we still won't fail if `hdiutil detach "#{@device}"` fails, as it still isn't the last command, which is possibly the command that is failing.

Can we implement the `set -e` feature we were discussing in #omnibus internal slack, or at least add a `set -e` to the top of every shellout block that is more than one command?

jaymalasinha:
@btm I have added the output progress tracking option available with hdiutil to avoid the resource busy option.  I think inbuilt command option will work better in this case.  Also I have added the dmg verification step to catch the 0bytes error.

jaymalasinha:
https://buildkite.com/chef/inspec-inspec-master-omnibus-adhoc/builds/54 passed so this is ready to merge

PR #896, opened 2019-08-02
Use architecture from platform mappings file when publishing packages

jeremiahsnapp:
### Description

This fixes the issue when in certain conditions packages published erroneously because the matching logic wasn't taking the architecture into account.

reference: https://github.com/chef/omnibus-buildkite-plugin/issues/22


scotthain:
Make sure you add the patch minor label @jeremiahsnapp

PR #915, opened 2019-11-14
Remove Solaris 10 platform and enable some hardening flags on Linux/Mac/FreeBSD

lamont-granquist:
We don't build or support Solaris 10 any more since it is EOL

This adds `-D_FORTIFY_SOURCE=2 -fstack-protector` to Linux/Mac/BSD.  Tried to add it to Solaris and it failed.  AIX is not gcc so its not applicable, and Windows mostly terrifies me.

This is likely a NOP since most distros have added this anyway, but it establishes a baseline.

This passed tests here:

https://buildkite.com/chef/chef-chef-master-omnibus-adhoc/builds/133

On 15.5.2 on Ubuntu we already have these settings:

```
# hardening-check /opt/chef/embedded/bin/openssl
/opt/chef/embedded/bin/openssl:
 Position Independent Executable: no, normal executable!
 Stack protected: yes
 Fortify Source functions: yes (some protected functions found)
 Read-only relocations: yes
 Immediate binding: no, not found!
```

This might do something useful like enable hardening on RHEL6 but I didn't exhaustively check.

### DIFF
--- a/lib/omnibus/software.rb
+++ b/lib/omnibus/software.rb
@@ -668,33 +668,22 @@ module Omnibus
             "ARFLAGS" => "-X64 cru",
           }
         when "solaris2"
-          if platform_version.satisfies?("<= 5.10")
-            solaris_flags = {
-              # this override is due to a bug in libtool documented here:
-              # http://lists.gnu.org/archive/html/bug-libtool/2005-10/msg00004.html
-              "CC" => "gcc -static-libgcc",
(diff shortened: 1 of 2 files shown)


marcparadise:
https://idea.popcount.org/2013-08-15-fortify_source/ was a helpful reference, I wasn't familiar with the effects of those flags .

lamont-granquist:
dang you're fast

PR #922, opened 2020-01-02
Fix call to URI#open

jkohen:
The hash needs to be expanded. As-is, it is an error with Ruby 2.7.0:

```
lib/omnibus/download_helpers.rb:80: warning: Splitting the last argument into positional and keyword parameters is deprecated
lib/omnibus/core_extensions/open_uri.rb:51:in `open_uri'
/usr/local/rvm/rubies/ruby-2.7.0/lib/ruby/2.7.0/open-uri.rb:153:in `open_uri': extra arguments (ArgumentError)
```

### Description

#### Maintainers

  serial number

### DIFF
--- a/lib/omnibus/download_helpers.rb
+++ b/lib/omnibus/download_helpers.rb
@@ -77,7 +77,7 @@ module Omnibus
           end
         end
 
-        file = open(from_url, options)
+        file = open(from_url, **options)
         # This is a temporary file. Close and flush it before attempting to copy
         # it over.
         file.close


jkohen:
I'll change this into a bug report. open() has a weird API that takes both symbols and string keys and arguments, and `**options` drops the string keys. This weird behavior seems to be working as "intended".

I tried something like `file = URI.open(from_url, "Accept-Encoding": "identity", **options)`, which passes the first batch of tests, but then fails the second batch:
```
  55) Omnibus::NetFetcher#fetch_required? when the file is downloaded when the checksum is the same return false
      Failure/Error: original_open_uri(name, *rest, &block)
      
      ArgumentError:
        unrecognized option: Accept-Encoding
      # ./lib/omnibus/core_extensions/open_uri.rb:51:in `open_uri'
      # ./lib/omnibus/download_helpers.rb:103:in `download_file!'
      # ./lib/omnibus/fetchers/net_fetcher.rb:176:in `download'
      # ./lib/omnibus/fetchers/net_fetcher.rb:86:in `fetch'
      # ./spec/functional/fetchers/net_fetcher_spec.rb:27:in `block (3 levels) in <module:Omnibus>'
      # ./spec/support/output_helpers.rb:10:in `capture_stdout'
      # ./spec/functional/fetchers/net_fetcher_spec.rb:27:in `block (2 levels) in <module:Omnibus>'
      # ./spec/functional/fetchers/net_fetcher_spec.rb:47:in `block (4 levels) in <module:Omnibus>'
```

PR #924, opened 2020-01-14
Implement deep signing for macOS Notarization

jonsmorrow:
### Description

This change adds deep signing of libraries and binaries to the pakacking process for the pkg packager. It aso enables the hardened runtime for binaries. In order to deep sign we have to know where a software definition will install binaries and libraries. To facilite this lib_dirs and bin_dirs functions have been added to the software definition. These two functions return defaults that match standard omnibus locations, but allow individual software defs to override if they do something different.

Bumps major version since signing deep signing is a major change.

Signed-off-by: Jon Morrow <jmorrow@chef.io>

#### Maintainers

  serial number


lib/omnibus/packagers/pkg.rb
```diff
@@ -177,6 +179,66 @@ def write_scripts
...
+          log.debug(log_key) { "    Signing: #{library}" }
+          sign_library(library)
+        end
+
+        log.info(log_key) { "  Signing binaries:" } unless binaries.empty?
+        binaries.each do |binary|
+          log.debug(log_key) { "    Signing: #{binary}" }
+          sign_binary(binary, true)
+        end
+        libraries.merge binaries
```
tyler-ball: What is the reason for merging the binaries into the libraries here after already processing the libraries?

jonsmorrow: Nothing specific. In one iteration of the code I think i may have been logging/displaying the list of things we signed. I guess I left it as a way for the caller to determine if anything was signed.

PR #935, opened 2020-03-03
Fix Ubuntu version in generated README.md

stevendanna:
The `ubuntu-1204` kitchen is no longer available in our default
generated omnibus project. This fixes the README we generate to
account for that.

Also, now we source the omnibus toolchain to make sure we have access
to bundler.

### DIFF
--- a/lib/omnibus/generator_files/README.md.erb
+++ b/lib/omnibus/generator_files/README.md.erb
@@ -47,7 +47,7 @@ $ bin/omnibus clean <%= config[:name] %> --purge
 ### Publish
 
 Omnibus has a built-in mechanism for releasing to a variety of "backends", such
-as Amazon S3. You must set the proper credentials in your 
+as Amazon S3. You must set the proper credentials in your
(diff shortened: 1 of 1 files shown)


lib/omnibus/generator_files/README.md.erb
```diff
@@ -105,17 +105,17 @@ individual build environment using the `kitchen` command.
...
 
 Then login to the instance and build the project as described in the Usage
 section:
 
 ```shell
-$ bundle exec kitchen login ubuntu-1204
+$ bin/kitchen login ubuntu-1804
+[vagrant@ubuntu...] $ .  load-omnibus-toolchain.sh
 [vagrant@ubuntu...] $ cd <%= config[:name] %>
 [vagrant@ubuntu...] $ bundle install
```
stevendanna: I think this step is still likely to fail because for some reason we are getting root owned files in `/home/vagrant/.bundle/cache/compact_index/`. My guess is something in the omnibus cookbook is doing this during provisioning, but I haven't dug in further than that.

tas50:
Progress

Issue #923, opened 2020-01-02
Ruby 2.7 failure in download_helpers


jkohen:
### Description

The hash needs to be expanded. As-is, it is an error with Ruby 2.7.0:

```
lib/omnibus/download_helpers.rb:80: warning: Splitting the last argument into positional and keyword parameters is deprecated
lib/omnibus/core_extensions/open_uri.rb:51:in `open_uri'
/usr/local/rvm/rubies/ruby-2.7.0/lib/ruby/2.7.0/open-uri.rb:153:in `open_uri': extra arguments (ArgumentError)
```

I tried to fix this in https://github.com/chef/omnibus/pull/922, but I ran into issues that I don't know how to fix, because I'm not very familiar with Ruby.

The Ruby documentation on the new behavior is comprehensive, but as far as I can see, there is no code that works in both Ruby versions without changing how things are wired: https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

### Omnibus Version

6.1.9

### Platform Version

Tell us which Operating System distribution, version and architecture the Omnibus build is being executed on.

### Replication Case

Tell us what steps to take to replicate your problem.  See [How to create a Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve)
for information on how to create a good replication case.

### Build Output

The relevant output of the Omnibus build or a link to a gist of the entire build, if there is one.

The verbose build output (`omnibus build <PROJECT> -l internal`) may be useful, but please link to a gist, or truncate it.

--------------------------------------------------

mattray:
Started doing Chef 16 builds with Ruby 2.7.1 and ran into this error, reverting to Ruby 2.6.6 avoids it. Here's a bigger stack-trace.

```
#<Thread:0x000055d56492cea0 /home/omnibus/omnibus-toolchain/.bundle/ruby/2.7.0/bundler/gems/omnibus-3327abfbe926/lib/omnibus/thread_pool.rb:57 run> terminated with exception (report_on_exception is true):
Traceback (most recent call last):
	14: from /home/omnibus/omnibus-toolchain/.bundle/ruby/2.7.0/bundler/gems/omnibus-3327abfbe926/lib/omnibus/thread_pool.rb:61:in `block (2 levels) in initialize'
[...]

PR #940, opened 2020-04-29
Ruby 2.7.1 adds a warning on calling open directly, needs to be URI.open

mattray:
Fixes https://github.com/chef/omnibus/issues/923
https://github.com/ruby/ruby/commit/3816cd945d68eac7ca8fecbc9d71f878ff3e7b3d#diff-0f19cb99597e5fb90bfb937b22143b51

### DIFF
--- a/lib/omnibus/download_helpers.rb
+++ b/lib/omnibus/download_helpers.rb
@@ -77,7 +77,7 @@ module Omnibus
           end
         end
 
-        file = open(from_url, options)
+        file = URI.open(from_url, options)
         # This is a temporary file. Close and flush it before attempting to copy
         # it over.
         file.close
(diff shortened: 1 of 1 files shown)


tas50:
This seems to cause a bunch of spec failures now

mattray:
Looking through the commits in https://github.com/ruby/ruby/commits/master/lib/open-uri.rb it looks like `URI.open` won't work before 2.7 and `open` won't work in 2.7 and later. I'm not much of a Rubyist, so I don't know how to check the Ruby version to provide conditional logic around this.

tas50:
@mattray this kind of thing:

```ruby
      if RUBY_VERSION.to_f < 2.7
        URI.encode(url)
      else
        CGI.escapeHTML(url)
      end
```

tas50:
Thanks @mattray

PR #951, opened 2020-06-18
Fill in some of the gotchas that are easy to hit with omnibus

jaymzh:
Fill in some of the gotchas that are easy to hit with omnibus

#### Maintainers

  serial number


jaymzh:
squashed them, together for DCO check.

jaymzh:
With these updated templates an instructions, I was able to make this work in kitchen pretty consistently and have now released several omnibus builds for SugarJar. This should be ready for merging.

tas50:
Thanks @jaymzh

PR #980, opened 2020-11-11
Replace chef-sugar for DSL helpers with chef-utils 15.4+

lamont-granquist:


spec/unit/software_spec.rb
```diff
@@ -207,12 +207,7 @@ module Omnibus
       end
 
       context "on aix" do
-        before do
-          # There's no AIX in Fauxhai :(
```
tas50, line 211: This actually isn't true. We have fauxhai data for aix

PR #981, opened 2020-11-13
Remove SLES 11 check

tas50:
SLES 11 is EOL. We don't need to perform this check at this point in
time.


PrajaktaPurohit (approved):
Thank you!

PR #983, opened 2020-11-13
Remove EOL Windows Support (pre-2012)

tas50:
We have some code checks in place that just don't need to run at this
point. 2008 R2 is EOL and it's not going to be built on at this point.


mwrock (approved):
Doesn't look like the ci failure is at all related to this PR.

tas50:
@mwrock Yep 2.7 has been broken here for a long time.

PR #999, opened 2021-02-11
Switch from -02 to -03 on Linux/macOS

tas50:
We're already doing -O3 on Windows


lamont-granquist (approved):

lamont-granquist:
there actually are specs on this one

PR #1006, opened 2021-03-12
Add Ruby 3 testing

tas50:

### DIFF
--- a/spec/unit/software_spec.rb
+++ b/spec/unit/software_spec.rb
@@ -190,8 +190,8 @@ module Omnibus
         end
       end
 
-      context "on macOS" do
-        before { stub_ohai(platform: "mac_os_x") }
+      context "on mac_os_x" do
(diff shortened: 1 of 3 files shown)


tas50:
Looks like this is going to need some work to get on 3

tas50:
@tduffield This is another good one msys folks could take on. This looks like it's another case of the argument changes in Ruby 3: https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

tduffield (approved):

spec/unit/software_spec.rb
```diff
@@ -190,8 +190,8 @@ module Omnibus
         end
       end
 
-      context "on macOS" do
-        before { stub_ohai(platform: "mac_os_x") }
```
tas50, line 194: Just FYI you don't need to specify the version here and if you leave it out then we don't need to update things in the future.

PR #1027, opened 2021-07-07
Updates rpm spec to use SHA256 for filedigest, from MD5

collinmcneese:
Signed-off-by: Collin McNeese <cmcneese@chef.io>

### Description

Updates the RPM spec for packaging to use SHA256 for the file digest algorithm instead of MD5.  This should allow support of RPM package installation on EL8-variant systems with FIPS-mode enabled.

Referenced in issue from https://github.com/chef/chef/issues/11790

thanks to @teknofire for the assistance in where the logic is running via omnibus

#### Maintainers

  serial number

### DIFF
--- a/resources/rpm/spec.erb
+++ b/resources/rpm/spec.erb
@@ -8,8 +8,8 @@
 %define __spec_clean_post true
 %define __spec_clean_pre true
 
-# Use md5
-%define _binary_filedigest_algorithm 1
+# Use SHA256 checksums for all files
+%define _binary_filedigest_algorithm 8
 
 %define _binary_payload <%= compression %>
 


tas50:
@collinmcneese there's a spec failure here

tas50 (approved):

tas50:
I suspect the reason we didn't do this originally was RHEL 5. It looks like it works fine on RHEL 6+. We'll see when we bring it into chef/chef.

collinmcneese:
> @collinmcneese there's a spec failure here

Yeah that was strange, there is no reason that should have failed, perhaps something transient with artifactory at the time

PR #1035, opened 2021-08-16
Replace usage of whitelist to allowlist in code, comments and DSL

balasankarc:
### Description

Instead of using `whitelist`, `whitelist_file`, etc. in code, comments and DSL, we can switch to `allowlist`, `allow_file` etc.

Note: This is a breaking change that would require a new major release. If a new major version is not on the horizon right now, we can drop the commit which changes` whitelist_file` to `allow_file` in this PR and do that later.

#### Maintainers

  serial number


lamont-granquist (approved):

tas50:
@balasankarc I want to make sure this gets done. I cut an issue to have one of our teams make sure we do this while also preserving the existing DSL. It would be great to throw warnings that these methods are deprecated. If you want to get this in quicker than our teams backlog feel free to update this PR with that. We can ship it as soon as we make sure the DSL doesn't immediately break.

PR #1078, opened 2022-08-30
Updated regex to match exactly 10, and 10.x

kanad-progress:
Signed-off-by: kanad-progress <kanad.kumar@progress.com>

### Description

#### Maintainers

  serial number

### DIFF
--- a/lib/omnibus/metadata.rb
+++ b/lib/omnibus/metadata.rb
@@ -218,7 +218,7 @@ module Omnibus
           # version is the same as Windows 2012R2. It's only here for completeness
           # and documentation.
           when /6\.3\.\d+/, "8.1" then "8.1"
-          when /^10\.0/ then "10"
+          when "10", /^10\.0/ then "10"
           else
             raise UnknownPlatformVersion.new(platform, platform_version)
           end


marcparadise:
Please update the commit message to describe what this addresses in words, instead of linking to the IPACK issue.  These issues have a tendency to vanish over the years.

lib/omnibus/metadata.rb
```diff
@@ -218,7 +218,7 @@ def truncate_platform_version(platform_version, platform)
           # version is the same as Windows 2012R2. It's only here for completeness
           # and documentation.
           when /6\.3\.\d+/, "8.1" then "8.1"
-          when /^10\.0/ then "10"
+          when /^10.*/ then "10"
           else
             raise UnknownPlatformVersion.new(platform, platform_version)
           end
```
marcparadise, line 224: To avoid making this catch to much, let's instead split it: 

`when "10", /^10\.0/ then "10"`

This way it will match exactly 10, and 10.x, but not "100".

marcparadise (changes requested):

marcparadise (approved):

vkarve-chef (approved):

PR #1083, opened 2022-10-05
INFC-268 fix Solaris build

tpowell-progress:
### Description

#### Maintainers

  serial number

### DIFF
--- a/lib/omnibus/health_check.rb
+++ b/lib/omnibus/health_check.rb
@@ -345,6 +345,12 @@ module Omnibus
       bad_libs = {}
       good_libs = {}
 
+      # The case/when below depends on the "current_library" being output with a : at the end
+      # and then the dependencies on subsequent lines in the form "library.so.1 => /lib/64/library.so.1"
+      # This output format only happens if ldd is passed multiple libraries (for Solaris, similar to Linux)
+      # FIXME if any of the `when` clauses in the `health_check_*` methods run before the `current_library`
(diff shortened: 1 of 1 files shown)


lib/omnibus/health_check.rb
```diff
@@ -345,11 +345,10 @@ def health_check_solaris
       bad_libs = {}
       good_libs = {}
 
-      read_shared_libs("find #{project.install_dir}/ -type f | xargs file | grep \"ELF\" | awk -F: '{print $1}' | sed -e 's/:$//'", "xargs -n 1 ldd") do |line|
+      read_shared_libs("find #{project.install_dir}/ -type f | xargs file | grep \"ELF\" | awk -F: '{print $1}' | sed -e 's/:$//'", "xargs ldd") do |line|
```
PrajaktaPurohit, line 348: could we add some description / context around the change and the breakage it is fixing?

tpowell-progress: Looking at something like

```ruby
      # The case/when below depends on the "current_library" being output with a : at the end
      # and then the dependencies on subsequent lines in the form "library.so.1 => /lib/64/library.so.1"
      # This output format only happens if ldd is passed multiple libraries (for Solaris, similar to Linux)
      # FIXME if any of the `when` clauses in the `health_check_*` methods run before the `current_library`
      # they probably should error out with an explicit callout of formatting with their environment's 
      # respective ldd parsing 
```

marcparadise (approved):

PR #1087, opened 2022-11-09
BS-94/use internal sources

gscho:
### Description

This is a new omnibus feature that allows users to build from only internal sources.

It works by adding a configuration value to the `omnibus.rb` file:

omnibus.rb
```ruby
use_internal_sources true
```

And then in the software definitions, you can define an `internal_source` like this:

my-software/cacerts.rb
```ruby
internal_source url: "https://my.internal.source/cacert-#{version}.pem"
```

When both source and internal_source are set, it will use the `internal_source` when `use_internal_sources` is set to true.

```ruby
source url: "https://curl.se/ca/cacert-#{version}.pem"
internal_source url: "https://my.internal.source/cacert-#{version}.pem" # uses this one
```

* When use_s3_caching is enabled, it will still use the cached sources wherever possible.
* When `use_internal_sources` is enabled and no `internal_source` is set for a dependency with a `url:` it will throw and exception

#### Maintainers

  serial number

### DIFF
--- a/lib/omnibus/config.rb
+++ b/lib/omnibus/config.rb
@@ -607,6 +607,12 @@ module Omnibus
     # @return [true, false]
     default(:health_check, true)
 
+    # Use the internal_source URL when fetching software dependencies. e.g. when creating an
+    # omnibus package using artifactory as the source url.
+    #
+    # @return [true, false]
+    default(:use_internal_sources, false)
+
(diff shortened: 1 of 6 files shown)


vkarve-chef (approved):
have tested this with error scenarios. works as expected.

PR #1135, opened 2023-11-20
Whitelist libnghttp2

ashiqueps:
### Description

To fix the chef-workstation vulnerabilities, we have added new versions of some of the packages on https://github.com/chef/omnibus-software/pull/1831. 

We have to whitelist libnghttp2 for the Mac OS as it was a dependency when we added Curl v8.4.0.

#### Maintainers

  serial number


ashiqueps:
This is no longer needed since we are adding the libnghttp2 as software in https://github.com/chef/omnibus-software/pull/1856

PR #1147, opened 2024-02-29
Add a small quality-of-life change for tests

deriamis:
### Description

In the Project and Library specs, we use an `expect` construct to ensure the build order is consistent. However, when there is a failure, we see a diff of the internal object representations. We can map the `:name` property into the construct instead to get a diff of the component names.

#### Maintainers

  serial number


spec/unit/library_spec.rb
```diff
@@ -113,7 +113,7 @@ def generate_software(name, version, dependencies = [])
         end
 
         it "returns an array of software descriptions, with all top level deps first, assuming they are not themselves transitive deps" do
-          expect(library.build_order).to eql(
+          expect(library.build_order(&:name)).to eql(
```
tpowell-progress, line 116: @deriamis should that be `build_order.map(&:name)` [again](https://github.com/chef/omnibus/pull/1147/files#diff-97f442dbeade17b123a0478ae9e6a7f7d482ac4e1399ffaf82891a7ec0523438R85), or am I missing something?

deriamis: No, you aren't missing anything. I am 🤡.

This is something I was moving over from something else I was doing, and I couldn't just cherry-pick it because it was part of something larger. Looks like I missed a couple of details. I'll push up a fix.

PR #1166, opened 2025-02-26
Fix `undefined method '=~' for false (NoMethodError)`

rgryta:
### Description

When providing a boolean override in the cli, omnibus will crash due to multi-staged casting.

#### Maintainers

  serial number


nilium:
Think this is a duplicate of #1160. Don't care which gets merged but hopefully the Chef folks eventually merge one of them so folks don't need to maintain patched omnibus forks.

